PR #1, opened 2023-06-28
[IOPLT-62] Adds all static const and types

CrisTofani:
## Short description
This PR includes the basic utils and variables to setup the new design system project


src/utils/variables.ts
dmnplb, line 1: A modest suggestion: why don't we delete them? Theoretically, all the new components don't reference these values. If we delete them, we can also quickly detect the technical debt because we have errors in the specific components.

dmnplb (changes requested):

src/utils/variables.ts
CrisTofani: [13a1929](https://github.com/pagopa/io-app-design-system/pull/1/commits/13a192977bdcd25e6c3b25541dbb24c1a59e0065) as suggested, may you check anything's ok?

PR #2, opened 2023-06-30
[IOPLT-63] Adds spacers

drmarro:
## Short description
Migrating spacers from io-app.

The following components have been migrated:
- Spacer
- Divider
- ComponentWrapper


src/utils/constants.ts
dmnplb, line 1: What's the advantage of having all these constants in a separate file? For example, the key `divider-default` is suggested by auto-complete and it's already present in the `IOTheme` object. If we keep it in a constant, do we lose this advantage?

drmarro: As general approach, having reused strings saved as constants could be more efficient and improves the code maintainability and readability.
In this case, I have seen that the two strings "horizontal" and "vertical" are reused multiple times in the io-app code so I have just reported them as constants in order to reuse them without rewriting any other string.
For the "divider-default", I agree with you: it is used just few times and maybe there's no reason to report it as constants.
What do you think about the horizontal and vertical constants ?

dmnplb: > What do you think about the horizontal and vertical constants ?

Before I answer, I'd like to understand what kind of problems we're preventing by using these kind of constants. If we're just using them to avoid two/three repetitions, maybe we're over-engineering the architecture.

drmarro: ok, I've reverted the change about the constants

dmnplb (approved):

PR #21, opened 2023-07-14
[IOPLT-90] Creating colors showcase

drmarro:
## Short description
Adding the showcase for the colors.

- Colors


example/src/pages/Colors.tsx
```diff
@@ -0,0 +1,370 @@
+import { H2, H3, H5, IOColorGradients, IOColors, IOColorsExtra, IOColorsLegacy, IOColorsNeutral, IOColorsStatus, IOColorsTints, IOStyles, IOThemeDark, IOThemeLight, LabelSmall, VSpacer, hexToRgba, themeStatusColorsDarkMode, themeStatusColorsLightMode } from "@pagopa/io-app-design-system";
```
dmnplb, line 1: Does it make sense to separate these import in different logic groups?

Here's my suggestion, but not 100% sure about it:

```suggestion
import { H2, H3, H5, LabelSmall } from "@pagopa/io-app-design-system/typography"
import { IOColorGradients, IOColors, IOColorsExtra, IOColorsLegacy, IOColorsNeutral, IOColorsStatus, IOColorsTints, hexToRgba, IOThemeDark, IOThemeLight, themeStatusColorsDarkMode, themeStatusColorsLightMode } from "@pagopa/io-app-design-system/color";
import { VSpacer } from "@pagopa/io-app-design-system/layout"
import { IOStyles } from "@pagopa/io-app-design-system"
```

drmarro: Ok, I've organized the imports in a different way.
Take a look to the new commit.
Thanks

dmnplb: @drmarro Is it still possible to import all the components the old way? I ask because the above example is quite extreme (and only exists on the DS color screen).

drmarro: Yes, it is possible.
The way the Design System components can be imported are defined here in the tsconfig.json -> https://github.com/pagopa/io-app-design-system/pull/21/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0

But it can be configured in order to force the user to separate logically the imports (import * from @pagopa/io-app-design-system/typography, import * from @pagopa/io-app-design-system/color) instead of having the possibility to import all from @pagopa/io-app-design-system.

CrisTofani: Are we sure the import are computed right by the app at run-time? At the beginning while developing and defining the paths in the `tsconfig.json` it looked to properly work at compile time but not at build time, did something changed?

drmarro: Yes @CrisTofani  it works only at build time but not at work time.
I've reverted the change.

CrisTofani: we will try to fix this behaviour in a future release

src/components/icons/index.tsx
```diff
@@ -1,2 +1,2 @@
 export * from "./Icon";
-export { Icon, AnimatedIcon } from "../icons/Icon";
+export { Icon, AnimatedIcon } from "./Icon";
```
CrisTofani, line 2: Is this needed? we already export all from `./Icon` in the previous line
```suggestion
```

CrisTofani (changes requested):

src/components/icons/index.tsx
drmarro: Yes, I'll remove the line.

CrisTofani (changes requested):

AleDore (approved):

PR #22, opened 2023-07-14
[IOPLT-91] Creating badge and tags showcase

drmarro:
## Short description
Creating showcase with Badge and Tags components.

DesignSystem:
- CustomBadge
- IOBadge
- PercentageValueBox

Example App:
- Badges


src/components/badge/IOBadge.tsx
```diff
@@ -0,0 +1,154 @@
...
+  },
+  labelSizeSmall: {
+    fontSize: 11,
+    lineHeight: 16
+  }
+});
+
+/**
+ * Official badge component
+@deprecated("Use the new Badge component instead")
```
dmnplb, line 108: This `Badge` component was developed as an official replacement for the legacy NativeBase Badge component. But visually it's the same as the legacy one. Are we sure to want to keep it in the new design system?

drmarro: Yes, you're right. 
I've removed the component and the Design System section.
Thanks for the hint.

PR #33, opened 2023-07-25
[IOPLT-110] Converting example app from expo project to react native one

drmarro:
## Short description
Converting the example app from expo project to a react native one.
This because there is a different way of handling fonts and (specifics) libraries from the io-app that can lead to confusion of usage.


ExampleApp/src/components/AssetViewerBox.tsx
```diff
@@ -0,0 +1,160 @@
+import { H5, IOColors, hexToRgba } from "@pagopa/io-app-design-system";
```
gunzip, line 1: what about adding a top level comment on each file to briefly explain its purpose?

CrisTofani (approved):
@drmarro PR LGTM, check out the latest changes made before merging 😄

CrisTofani (approved):
LGTM

PR #44, opened 2023-08-10
[IOPLT-113] Input text component

CrisTofani:
## Short description
This PR implements the new input text component in the library

## How to test
Use the example app navigating through the section `Input Texts`

https://github.com/pagopa/io-app-design-system/assets/3959405/548f88a3-ca75-4167-8852-eba63c6921f1


src/components/textInput/TextInputBase.tsx
```diff
@@ -0,0 +1,348 @@
...
+            setInputStatus("focused");
+            setIsValid(undefined);
+            labelSharedValue.value = 1;
+          }}
+          maxLength={counterLimit}
+          onBlur={onBlurHandler}
+          value={value}
+          onChangeText={onChangeTextHandler}
+          style={[
+            onValidate || secretInput ? { width: "80%" } : {},
```
dmnplb, line 267: Setting an arbitrary value (or magic number, like `80%`) for the width will result in this kind of behavior:
Simulator Screen Shot - iPhone 14 Pro - 2023-08-22 at 16 02 09


To avoid the text overlapping the icon, we should build the style logic with flex and let it regulate the width depending on the context (with or without icon, etc...)

src/components/textInput/TextInputBase.tsx
```diff
@@ -0,0 +1,348 @@
...
+        />
+        <Animated.View
+          style={[styles.textInputLabelWrapper, icon ? { left: 32 } : {}]}
+        >
+          <Animated.Text
+            numberOfLines={1}
+            style={[
+              animatedLabelProps,
+              {
+                backgroundColor: IOColors.white,
```
dmnplb, line 279: Have you encountered any particular problems with setting the `backgroundColor` property? Because I tried to remove it and the only change was the height of the input caret (which will change anyway once the floating label ends)

src/components/textInput/TextInputBase.tsx
```diff
@@ -0,0 +1,348 @@
...
+    secretInput ?? false
+  );
+
+  const inputRef = useRef<TextInput>(null);
+
+  const boxStyle: ViewStyle = useMemo(() => {
+    if (inputStatus === "focused") {
+      return {
+        borderColor: IOColors["blueIO-500"],
+        borderWidth: 2
```
dmnplb, line 165: Changing the `borderWidth` on the focus state results in a slight visual glitch on the contained text (it moves a bit):

https://github.com/pagopa/io-app-design-system/assets/1255491/aeec7c40-1630-42d0-ac62-34fcd6714eb5

It's a minor issue, but I'd try to use `boxShadow` property instead (with zero blur, used as substitute of the border)

src/components/textInput/TextInputBase.tsx
```diff
@@ -0,0 +1,348 @@
...
+        onPress={onTextInputPress}
+        style={[
+          inputStatus === "disabled" ? { opacity: 0.5 } : {},
+          boxStyle,
+          styles.textInput
+        ]}
+      >
+        {icon && (
+          <>
+            <Icon name={icon} size={24} />
```
dmnplb, line 247: According to the design guidelines, the icon color is not dark:
```suggestion
            <Icon name={icon} color="grey-300" size={24} />
```

src/components/textInput/TextInputBase.tsx
```diff
@@ -0,0 +1,348 @@
...
+    }
+    return {
+      borderColor: IOColors["grey-200"],
+      borderWidth: 1
+    };
+  }, [inputStatus]);
+
+  const animatedLabelProps = useAnimatedStyle(() => ({
+    fontSize: withTiming(interpolate(labelSharedValue.value, [0, 1], [16, 12])),
+    top: withTiming(interpolate(labelSharedValue.value, [0, 1], [0, -16]))
```
dmnplb, line 182: I don't really like magic numbers without explanation, but:
- `16` is too close to the edge
- `12` is good, but it starts to overlap the input caret
```suggestion
    transform: [
      {
        translateY: withTiming(
          interpolate(labelSharedValue.value, [0, 1], [0, -14]),
          {
            duration: 300,
            easing: Easing.elastic(0.85)
          }
        )
      }
    ]
```
Also, I'm probably biased by the best practices about Web animation performance (maybe not valid in the RN layout engine), but applying the transition using `translateY` should be more performant and also clearer to the reader

dmnplb (commented):
After playing with them, I have some notes other than specific references to the code:

- The logic of the component is starting to look quite complex. Have you considered creating a reusable BaseInput (which manages the common transitions) and then creating different TextField components depending on the intended use? (e.g.: InputPassword, CounterField, etc...). In our main codebase we have the `LabelledItem` which manages everything and it seems unmaintainable to me
- Is there a way to center the password dots around the caret? (See video below)

https://github.com/pagopa/io-app-design-system/assets/1255491/ac3de905-f6e5-4c2b-86f5-c02ae1fd7f21

src/components/textInput/TextInputBase.tsx
CrisTofani: Unfortunately it's not possible to remove the blur from the shadow in RN, I just tested it on iOS but as far as I remember shadow actually acts in a different way between android and iOS

This is the max result i got, but it's far way different from the mock up in Figma.

src/components/textInput/TextInputBase.tsx
```diff
@@ -0,0 +1,352 @@
...
+  value = "",
+  icon,
+  counterLimit,
+  helperText,
+  onChangeText,
+  onValidate,
+  secretInput,
+  errorMessage
+}: InputTextProps) => {
+  const labelSharedValue = useSharedValue(0);
```
dmnplb, line 151:
```suggestion
  const labelSharedValue = useSharedValue<boolean>(false);
```
If `labelSharedValue` acts as boolean, we should specify it and then make the animation simpler with a condition, without `interpolation` method:

```
fontSize: withTiming(labelSharedValue.value ? 12 : 16, {
      duration: 300,
      easing: Easing.elastic(0.85)
    }),
```

CrisTofani:
@dmnplb completely reworked on component logic to split up the logic on Validation component and password component, may you please check latest commits?

dmnplb (approved):
Not everything is fixed yet, especially from a visual point of view, but we should release it as a first iteration and get some feedback from our colleagues :)

PR #53, opened 2023-08-30
[IOPLT-112] Adds the FooterWithButtons component

CrisTofani:
## Short description
This PR creates a new version of the footer with buttons with the new version of buttons from DS library

## How to test
Check the specific page in the Example app to see how it changes.


src/components/contentWrapper/ContentWrapper.tsx
```diff
@@ -19,6 +19,7 @@ export const ContentWrapper = ({
 }: IOContentWrapperProps) => (
   <View
     style={{
+      flex: 1,
```
dmnplb, line 22:
```suggestion
```
Tried to delete it and nothing changed. In general, we shouldn't change a layout prop in a generic component like `ContentWrapper` because it's quite risky.

dmnplb (changes requested):
Compared to the old one, I can't see the shadow applied to the main container. This absence could potentially change the visual appearance of the page interested by the component replacement. Did I miss something?

src/components/contentWrapper/ContentWrapper.tsx
CrisTofani: this was due to some testing, we can revert with no problem

CrisTofani:
@dmnplb shadow has bene restored in [1bdec95](https://github.com/pagopa/io-app-design-system/pull/53/commits/1bdec95ec1746b3eeac6483c4d096dcabda3088e)

dmnplb (approved):

PR #57, opened 2023-09-06
[IOPLT-141] Changes the Tag component to handle even the only icon or only text visualization

CrisTofani:
## Short description
This PR adds two new possible variants of Tag component:
- Only Icon
- Only Text

## How to test
To test check anything breaks the previous version of Tag component on the example app.


dmnplb (commented):
Just to make sure there's no extra margin when there's no text set, we should add the margin to the text instead of the icon:
```
iconWrapper: {
    flexShrink: 1
  },
label: {
    marginStart: IOTagIconMargin,
    fontSize: 12,
```
**EDIT**: The solution I gave isn't correct either. We should only use a `HSpacer` component when there's both text and icon.

src/components/tag/Tag.tsx
```diff
@@ -70,7 +84,6 @@ const styles = StyleSheet.create({
     paddingVertical: IOTagVSpacing
   },
   iconWrapper: {
-    marginEnd: IOTagIconMargin,
     flexShrink: 1
   },
```
dmnplb, line 88:
```suggestion
  },
  spacer: {
    width: IOTagIconMargin
  },
```

src/components/tag/Tag.tsx
```diff
@@ -89,15 +102,23 @@ const styles = StyleSheet.create({
...
+      O.fold(
+        () => null,
+        ({ iconColor, iconName }) => (
+          <View style={styles.iconWrapper}>
+            <Icon name={iconName} color={iconColor} size={IOTagIconSize} />
+          </View>
+        )
+      )
+    )}
+    {mapVariants[variant] && text && <HSpacer size={8} />}
```
dmnplb, line 117: Not a perfect implementation, but it's just an exception for this particular component. If other components start using this `6px` spacing value, we should consider adding it to the `V/HSpacer` spec.
```suggestion
    {mapVariants[variant] && text && <View style={styles.spacer} />}
```

dmnplb (changes requested):
Unfortunately, we have to be very strict about the spacing because it's a component used for the `MessageListItem` where every pixel counts.

src/components/tag/Tag.tsx
CrisTofani: What about adding the value 6 to the spacer?

dmnplb: According to the design specs, there is no `6` value in the spacing scale used for `H/VSpacer` components. Adopting it means adding that value to the `Layout` DS page, which gives the false impression that it's a widely accepted value. It's not, because it's only used here, for now.

dmnplb: The `6` value is already added to the `IOSpacingScale` type, btw:
```ts
const IOTagIconMargin: IOSpacingScale = 6;
```

dmnplb (approved):

PR #59, opened 2023-09-08
[IOPLT-144] Map Legacy buttons versions to the context attribute

CrisTofani:
# ⚠️ THIS PR DEPENDS ON #56

## Short description
This PR implements the handler to chenge UI representation of the button components to use the legacy view based on `IOExperimentalDesignContext`

## How to test
Check on the example app in the button section components changes on toggle value change

https://github.com/pagopa/io-app-design-system/assets/3959405/611bef78-4ff2-40d6-9978-62b623a9b4a3


dmnplb (changes requested):
Just a reminder about this legacy button context mapping activity. Our first goal is not to optimize the code, but rather to make the removal of the legacy code as easy and quick as possible. For example, if the component logic contains more than four/five conditions on the context value, we should consider splitting the code logic completely between the legacy and new variants, even at the cost of duplicate code.

Last but not least, we should restore the previous comments ("start/end legacy code") to make future removal even easier.

dmnplb (approved):

CrisTofani:
@dmnplb added in [c589fae](https://github.com/pagopa/io-app-design-system/pull/59/commits/c589fae41e16329681284c20d19a12283e62877d) some `TODO`s comments as a reference to remove all legacy code.

Talking about code optimization, I can't totally agree, since a good organized code may be easier to read even if a wide part of it should be deleted in a future step. To have a legacy view doesn't always mean to have wide code duplication, that's what I meant removing the legacy component file. 😄

PR #67, opened 2023-09-14
feat: [IOBP-241] Add `IconChange` and `IconDialpad` icons

Hantex9:
## Short description
This PR adds two new icons: `IconChange` and `IconDialpad` used [inside this bottom sheet page](https://www.figma.com/file/SuYjyjZlieQqRS8n1B6qbF/Flussi-App-ID-Pay?type=design&node-id=9611-30793&mode=design&t=mqcM2GWOFRs4RGrR-4).

- Added svgs original files and react components


dmnplb (commented):
I know it's a minor issue (I should probably add it to the local README about adding new icons), but to facilitate the fruition of the icons in the **_DS Icons_** page, we should add the icons near the conceptually similar ones. In this case the `change` key should be grouped with other arrow icons. Also `dialpad` near the `device` icon.

It's not mandatory, but it's nice to have.

(To check the order, just launch the `example` app to quickly look at the page)

Hantex9:
> I know it's a minor issue (I should probably add it to the local README about adding new icons), but to facilitate the fruition of the icons in the **_DS Icons_** page, we should add the icons near the conceptually similar ones. In this case the `change` key should be grouped with other arrow icons. Also `dialpad` near the `device` icon.
> 
> It's not mandatory, but it's nice to have.
> 
> (To check the order, just launch the `example` app to quickly look at the page)

Alright, I agree with you about keeping icons grouped together! Addressed into https://github.com/pagopa/io-app-design-system/pull/67/commits/21e63c4eab89780cae3e328e62979bb3bf8a5e87

dmnplb (approved):

PR #77, opened 2023-09-20
[IOPLT-135] Refinement of the `HeaderSecondLevel` Component

CrisTofani:
## Short description
This PR refines the component definition of the Header of the second level navigation pages.

## How to test
Check the related Header Second Level example app page to verify everything's fine


dmnplb (commented):
Does it make sense to include an example of the 2nd level header without animation? Also to understand how it behaves without adding a `ScrollView` with `triggerOffsetValue`?

dmnplb (approved):

PR #78, opened 2023-09-21
[IOPLT-138] Creates the new ModulePaymentNotice

CrisTofani:
## Short description
This PR introduces the new ModulePaymentNotice in the library components

- new ModulePaymentNotice component

## How to test
Check the related Section in the Modules screen of the example app

### DIFF
--- a/src/components/typography/LabelSmall.tsx
+++ b/src/components/typography/LabelSmall.tsx
@@ -1,6 +1,6 @@
 import { useIOExperimentalDesign } from "../../core";
 import type { IOColors, IOTheme } from "../../core/IOColors";
-import { FontFamily, IOFontFamily, IOFontWeight } from "../../utils/fonts";
+import { FontFamily, IOFontWeight } from "../../utils/fonts";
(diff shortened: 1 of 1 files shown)


dmnplb (changes requested):
I just noticed a couple of visual inconsistencies with the experimental DS set to "on", especially when compared to the design artboards that came with it:
- The arrow icon has the old blue instead of the new one.
- The title of the payment notice is set with Titilium, instead of Readex.

CrisTofani:
> * The title of the payment notice is set with Titilium, instead of Readex.

@dmnplb Title is rendered through `LabelSmall` component which has no `ReadexPro` variant in the new DS components (or we missed to implement it) do you have any other suggested component to use in this case?

dmnplb:
> Title is rendered through LabelSmall component which has no ReadexPro variant in the new DS components (or we missed to implement it) do you have any other suggested component to use in this case?

We probably could use the opportunity to add the missing typographic style (that's transformed to Titilium with DS set to off). What do you think?

PR #79, opened 2023-09-21
feat: [IOBP-236] ListItemSwitch extension with loading state and badge 

Hantex9:
## Short description
This PR adds three more props to `ListItemSwitch` component:
- `isLoading`: while true it shows an activity indicator
- `badgeText`: A text to show inside a badge instead of the switch. If it is not null, you have to provide also a `badgeVariant` 
- `badgeVariant`: It indicates the badge variant to show;

- Added three new props, the `badgeText` and `badgeVariant` are conditional properties;
- Changed the position of the right component and aligned to the title;
- Added more examples about `ListItemSwitch` inside example app;

## How to test
Open the example app, navigate through `List items` section and check the new examples:

## Preview
https://github.com/pagopa/io-app-design-system/assets/34343582/bd154d69-b7af-44ac-b09d-34674c80ab9e


src/components/listitems/ListItemSwitch.tsx
```diff
@@ -10,14 +15,25 @@ import { IOIcons, Icon } from "../icons";
...
-};
+  isLoading?: boolean;
+} & (
+  | {
+      badgeText: string;
+      badgeVariant: Pick<Badge, "variant">["variant"];
+    }
+  | {
+      badgeText?: undefined;
+      badgeVariant?: undefined;
```
CrisTofani, line 34:
```suggestion
      badgeText?: never;
      badgeVariant?: never;
```

what about using type `never` in this case?

CrisTofani (changes requested):

src/components/listitems/ListItemSwitch.tsx
Hantex9: That's absolutely better! Addressed in https://github.com/pagopa/io-app-design-system/pull/79/commits/35235694105bde937655ca7f0c2e3e2214d98813

CrisTofani:
@dmnplb I have doubt about this implementation: the badge version of the list item should be explicit of this specific list item or be took apart as a scoped component?

Hantex9:
> @dmnplb I have doubt about this implementation: the badge version of the list item should be explicit of this specific list item or be took apart as a scoped component?

For any doubt about this component, [this is the figma](https://www.figma.com/file/SuYjyjZlieQqRS8n1B6qbF/Flussi-App-ID-Pay?type=design&node-id=10086-13063&mode=design&t=h9zc4GHhqNPPPnbh-4) where this component is used. I think that is overkill to create another component only to handle a badge on the right side and it will have negative effect on the DX

CrisTofani:
> > @dmnplb I have doubt about this implementation: the badge version of the list item should be explicit of this specific list item or be took apart as a scoped component?
> 
> For any doubts about this component, [this is the figma](https://www.figma.com/file/SuYjyjZlieQqRS8n1B6qbF/Flussi-App-ID-Pay?type=design&node-id=10086-13063&mode=design&t=h9zc4GHhqNPPPnbh-4) where this component is used. I think that is overkill to create another component only to handle a badge on the right side and it will have negative effect on the DX

Looking to the figma you linked it could even be represented by another component (eg: A standard ListItem with a badge on the right) it doesn't look to be exclusively a ListItemSwitch with the badge rendered conditionally to the Switch element.

I would prefer to keep the ListItemSwitch closed to it's purpose

dmnplb:
> Looking to the figma you linked it could even be represented by another component (eg: A standard ListItem with a badge on the right) it doesn't look to be exclusively a ListItemSwitch with the badge rendered conditionally to the Switch element.

Unfortunately, we can't use this approach because a simple `ListItem` does not look like a `ListItemSwitch` without a badge. It has different styles for title and description. In this particular case, we want to have consistent styles between the different cases when we compose them on a single page.

So for me, extending the `ListItemSwitch` may be a viable option. Fortunately, we don't have to manage the pressed behavior for the entire block like other **Selection** components.

src/components/listitems/ListItemSwitch.tsx
```diff
@@ -10,14 +15,25 @@ import { IOIcons, Icon } from "../icons";
...
   onSwitchValueChange?: (newValue: boolean) => void;
   description?: string;
   icon?: IOIcons;
   action?: SwitchAction;
-};
+  isLoading?: boolean;
+} & (
+  | {
+      badgeText: string;
+      badgeVariant: Pick<Badge, "variant">["variant"];
```
dmnplb, line 30: What about using an `object` instead of two different props, like the `action` above? So the code below could be slightly clearer (`badge` instead of `badgeText`):

```jsx
{!isLoading && !badge && (
```

CrisTofani: I would suggest to keep the strong handling for the typedef.

Something like this could be accurate:
```typescript
{
  ...rest
} & (
  | {
      badge: {
        badgeText: string;
        badgeVariant: Pick<Badge, "variant">["variant"];
      };
    }
  | {
      badge?: never;
    }
)
```

Hantex9: What do you think about the implementation addressed into https://github.com/pagopa/io-app-design-system/pull/79/commits/91c4bddee31b2f61f06d7d8a2a378f1130c601dc?

Since the `Badge` type have only 2 required props (text and variant) and an optional property that is the testID (that could be useful in some test cases), in this case there is no more to handle the conditional typedef. What do you think?

CrisTofani (approved):
lgtm

dmnplb (changes requested):
I'll approve it when the components are located in the right section of the example playground (under **Selection**, instead of **List Items**)

Hantex9:
> I'll approve it when the components are located in the right section of the example playground (under **Selection**, instead of **List Items**)

The suggested hint is already addressed into commit 71b218315c6714cb85a432945588b7864baa0e89 on Friday, please take a look at it

dmnplb:
@Hantex9 My bad, I forgot to pull the latest updates 😅

dmnplb (approved):

PR #91, opened 2023-10-02
[IABT-1490] Fix amount text a11y in `ListItemTransaction`

mastro993:
## Short description
This PR fixes an accessibility issue with the `ListItemTransaction` where, on iOS, VoiceOver reads time and amount numbers as one number (eg: `23:15 45,67€` reads "23" and "1.545,67€").

The fix consists in setting to `false` the `accessible` prop in the parent component. Now the `accessible` prop is passed from the parent component, default is `false`.

- [src/components/listitems/ListItemTransaction.tsx](https://github.com/pagopa/io-app-design-system/pull/91/files#diff-e161c812707e7a718a116a2c9a28e50d9d46a9956ffa18d7950aafe270570ed9): added the `accessible` prop in the parent component.

## How to test
Use VoiceOver on iOS to read an `ListItemTransaction` component.
This issue should be tests on physical devices since it cannot be reproduced on simulators.

### DIFF
--- a/src/components/listitems/ListItemTransaction.tsx
+++ b/src/components/listitems/ListItemTransaction.tsx
@@ -64,6 +64,7 @@ export type ListItemTransaction = WithTestID<
     paymentLogoIcon?: ListItemTransactionLogo;
     subtitle: string;
     title: string;
+    accessible?: boolean;
   } & (
       | {
           transactionStatus: ListItemTransactionStatusWithoutBadge;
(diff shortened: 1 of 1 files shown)


dmnplb:
The solution works, but seems a bit hacky. I think we can quickly approve it as a temporary solution, but isn't there a way to mark the content as "time value"? From the official React Native docs it seems not 🤔

dmnplb (approved):

mastro993:
> The solution works, but seems a bit hacky. I think we can quickly approve it as a temporary solution, but isn't there a way to mark the content as "time value"? From the official React Native docs it seems not 🤔

It seems like all the contents of the ListItem are being sent to VoiceOver as one string and, for some reason, it can't figure out how to read it. I discovered that this only happens with the Italian language.
We should open a ticket to investigate further later, but for now this hack seems to be working pretty well.

CrisTofani (approved):

PR #94, opened 2023-10-04
[IOAPPFD0-168] Enable font scaling for the `Button…` components

dmnplb:
## Short description
This PR enables font scaling for the `Button...` components only when the experimental DS is enabled.

- Enable font scaling for the `ButtonSolid`, `ButtonOutline` and `ButtonLink` components

## How to test
1. Launch the example app
2. Go to the **Buttons** page
3. Change dynamic type values with the Accessibility inspector


src/components/typography/BaseTypography.tsx
```diff
@@ -54,7 +54,7 @@ export const BaseTypography: React.FC<OwnProps> = props => {
     : [props.fontStyle, fontStyle];
 
   return (
-    <Text {...props} style={style}>
+    <Text allowFontScaling={false} {...props} style={style}>
```
CrisTofani, line 57: what about wiring this to the experimental context too?

dmnplb: Note that `io-app` has a default prop with `allowFontScaling` set to false. So it's kind of a redundant prop for proper testing in the `example` app.

CrisTofani (approved):

CaraDune:
what a time to be alive <3 thanks guys, let's keep up the good work

PR #97, opened 2023-10-06
[IOAPPFD0-171] Add the new `LoadingSpinner` component

dmnplb:
## Short description
This PR adds the new `LoadingSpinner` component, derived from `ItwLoadingSpinner` developed specifically for the IT Wallet flow. To learn more about the original component:
* https://github.com/pagopa/io-app/pull/5055

- Slightly refactor the original component
- Add the new **Loaders** page to the example app

### Preview

https://github.com/pagopa/io-app-design-system/assets/1255491/fa506c48-eea3-4210-854c-b5fe66ba5132

## How to test
1. Launch the example app
2. Go to the **Loaders** page


LazyAfternoons:
I just want to leave a comment regarding this loader. The version implemented here is temporary as the actual loading indicator has a gradient:
Screenshot 2023-10-06 at 15 20 35

dmnplb:
@LazyAfternoons Yes, I'm aware of the difference. But we could still release it with the prospect of refining it later. Also we're going to change the following things:
- Leave only one prop (`foregroundColor` or just `color`) without the need to set the `backgroundColor`
- Probably rewrite using Reanimated instead of Animated (low priority)

dmnplb:
@LazyAfternoons Improved it by making it look like the designed one. It doesn't use the border hack anymore, but everything is made with a custom SVG. To summarize everything:
- there is only one `color` prop, so we don't need to set `backgroundColor` anymore
- you can also set a custom `stroke` value

@CrisTofani Fixed the TS issue with the new page by deleting it. YOLO.

LazyAfternoons:
> @LazyAfternoons Improved it by making it look like the designed one. It doesn't use the border hack anymore, but everything is made with a custom SVG. To summarize everything:
> 
> * there is only one `color` prop, so we don't need to set `backgroundColor` anymore
> * you can also set a custom `stroke` value
> 
> @CrisTofani Fixed the TS issue with the new page by deleting it. YOLO.

Thank you, it looks good to me. I think the `Loaders` page is missing from the example app though.

dmnplb:
>  I think the Loaders page is missing from the example app though. Is it missing due to the TS issue you mentioned?

Yes, it's missing because we have a TS issue when we add the new page. But it's a minor issue because the official DS page will be in the official repository

CrisTofani (approved):

PR #99, opened 2023-10-09
[IOAPPFD0-172] Add `loading` state to the `ButtonSolid` (with custom transitions)

dmnplb:
## Short description
Include a summary of the changes.

- Feature A
- Feature B

## How to test
Describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

### DIFF
--- a/src/components/buttons/ButtonSolid.tsx
+++ b/src/components/buttons/ButtonSolid.tsx
@@ -11,6 +11,7 @@ import Animated, {
   withSpring,
   withTiming
 } from "react-native-reanimated";
+import ReactNativeHapticFeedback from "react-native-haptic-feedback";
 import { IOIconSizeScale, IOIcons, Icon } from "../icons";
 import { WithTestID } from "../../utils/types";
 import { HSpacer } from "../spacer/Spacer";
@@ -299,6 +300,11 @@ export const ButtonSolid = React.memo(
       isPressed.value = 0;
     }, [isPressed]);
 
+    const handleOnPress = (event: GestureResponderEvent) => {
+      ReactNativeHapticFeedback.trigger("impactLight");
+      onPress(event);
+    };
+
     // Label & Icons colors
     const foregroundColor: IOColors = disabled
       ? colorMap[color]?.label?.disabled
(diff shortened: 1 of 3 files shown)


src/components/buttons/ButtonSolid.tsx
```diff
@@ -211,29 +221,35 @@ export const ButtonSolid = React.memo(
...
         accessibilityHint={accessibilityHint}
+        accessibilityState={{ busy: loading }}
         accessibilityRole={"button"}
-        testID={testID}
-        onPress={onPress}
+        onPress={handleOnPress}
         onPressIn={onPressIn}
         onPressOut={onPressOut}
-        accessible={true}
         disabled={disabled}
```
CrisTofani, line 245: maybe we should force a disabled state while the loading is running, in this way we may prevent from any other Tap by the user, or at least if we don't need to alter the UI of the button maybe we can _disable_ the onPress event in some way.

What do you think?

CrisTofani (changes requested):

src/components/buttons/ButtonSolid.tsx
dmnplb: Tried to add it, but the addition disabled the scale effect applied to the Button (even when it's not in loading state)

dmnplb: @CrisTofani Added a new line to avoid calling `onPress`

CrisTofani: What about instead changing the handleOnPress with a useCallback? 

Something like:

```typescript
    const handleOnPress = useCallback((event: GestureResponderEvent) => {
      if(isLoading) {
        return;
      }
      ReactNativeHapticFeedback.trigger("impactLight");
      onPress(event);
    }, [isLoading, onPress]);
```

dmnplb: Seems cleaner. Addressed in the following commit https://github.com/pagopa/io-app-design-system/pull/99/commits/09bcb189f1364302b5a6ab9e022b0691dcd90f54

CrisTofani (approved):

PR #111, opened 2023-10-19
chore: [IOBP-352] Add paymentLogo prop into `ListItemNav`

Hantex9:
## Short description
This PR introduces a new `ListItemNav` prop: `paymentLogo` which allows to show an icon on the left side which is a payment logo instead of IOIcons.

- Added a new prop `paymentLogo` into component

## How to test
Open the example app and navigate to "List item" section and you should see a list item nav with a bancomat pay logo on the left

## Preview
image

### DIFF
--- a/src/components/listitems/ListItemNav.tsx
+++ b/src/components/listitems/ListItemNav.tsx
@@ -20,7 +20,6 @@ import {
   IOListItemStyles,
   IOListItemVisualParams,
   IOScaleValues,
-  IOSelectionListItemVisualParams,
   IOSpringValues,
   IOStyles,
   hexToRgba,
@@ -43,7 +42,7 @@ const legacyStyles = StyleSheet.create({
   }
 });
 
-export type ListItemNavPartialProps = WithTestID<{
+type ListItemNavPartialProps = WithTestID<{
   value: string | React.ReactNode;
   description?: string | React.ReactNode;
   onPress: (event: GestureResponderEvent) => void;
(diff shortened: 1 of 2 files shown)


src/components/listitems/ListItemNav.tsx
```diff
@@ -171,6 +178,14 @@ export const ListItemNav = ({
               />
             </View>
           )}
+          {paymentLogo && (
+            <View style={{ marginRight: IOListItemVisualParams.iconMargin }}>
+              <LogoPayment
+                name={paymentLogo}
+                size={IOSelectionListItemVisualParams.iconSize}
```
dmnplb, line 185: Stick to the `IOListItemVisualParams` object, instead of importing the one related to **Selection** components
```suggestion
                size={IOListItemVisualParams.iconSize}
```

src/components/listitems/ListItemNav.tsx
```diff
@@ -41,20 +43,27 @@ const legacyStyles = StyleSheet.create({
   }
 });
 
-export type ListItemNav = WithTestID<{
+export type ListItemNavPartialProps = WithTestID<{
```
dmnplb, line 46: Maybe it's better to not export the `PartialProps` type:
```suggestion
type ListItemNavPartialProps = WithTestID<{
```

dmnplb (changes requested):

src/components/listitems/ListItemNav.tsx
Hantex9: Addressed into 48ff0bbd8c7d54c311814065e18c302e6c61a6a6

src/components/listitems/ListItemNav.tsx
Hantex9: Addressed into 48ff0bbd8c7d54c311814065e18c302e6c61a6a6

dmnplb (approved):

PR #126, opened 2023-11-02
[IOPLT-230] GradientBottomActions refinement 

drmarro:
## Short description
Refining GradientBottomActions component. It actually has two generic parameters (firstAction and secondAction): these two params has to be specific components (ButtonSolid or ButtonLink).

- GradientBottomActions
- ButtonSolid
- ButtonLInk


src/components/layout/GradientBottomActions.tsx
```diff
@@ -99,7 +101,12 @@ export const GradientBottomActions = ({
       ]}
     />
     <View style={styles.buttonContainer} pointerEvents="box-none">
-      {primaryAction}
+      {primaryAction &&
+        (primaryAction.type === "ButtonLink" ? (
```
dmnplb, line 105: We don't need this condition because the primary action is always a `ButtonSolid` and the secondary action is always a `ButtonLink`.

src/components/layout/GradientBottomActions.tsx
```diff
@@ -109,7 +116,11 @@ export const GradientBottomActions = ({
           }}
         >
           <VSpacer size={dimensions.spaceBetweenActions} />
-          {secondaryAction}
+          {secondaryAction.type === "ButtonLink" ? (
```
dmnplb, line 119: Same as above

src/components/layout/common.ts
dmnplb, line 1: Given the requirements specified in the comments of the `GradientScrollView` file, do we really need this logic? Why we can't keep it in the main file?

src/components/layout/GradientBottomActions.tsx
```diff
@@ -1,18 +1,20 @@
...
 
 export type GradientBottomActions = WithTestID<{
   transitionAnimStyle: Animated.AnimateStyle<StyleProp<ViewStyle>>;
   dimensions: GradientBottomActionsDimensions;
   // Accepted components: ButtonSolid, ButtonLink
   // Don't use any components other than this, please.
-  primaryAction?: React.ReactNode;
-  secondaryAction?: React.ReactNode;
+  primaryAction?: GradientBottomAction;
+  secondaryAction?: GradientBottomAction;
```
dmnplb, line 17: We can just set the relative types, without being generic:
```suggestion
  primaryAction?: ButtonSolidAction;
  secondaryAction?: ButtonLinkAction;
```

src/components/layout/GradientScrollView.tsx
```diff
@@ -8,23 +8,24 @@ import Animated, {
...
 export type GradientScrollView = WithTestID<{
   children: React.ReactNode;
   excludeSafeAreaMargins?: boolean;
   debugMode?: boolean;
   // Accepted components: ButtonSolid, ButtonLink
   // Don't use any components other than this, please.
-  primaryAction: React.ReactNode;
-  secondaryAction?: React.ReactNode;
+  primaryAction: GradientBottomAction;
+  secondaryAction?: GradientBottomAction;
```
dmnplb, line 28:
```suggestion
  primaryAction: ButtonSolidAction;
  secondaryAction?: ButtonLinkAction;
```

dmnplb (changes requested):

src/components/layout/GradientBottomActions.tsx
drmarro: Ah ok. I believe that both the primary and the secondary action could be a ButtonSolid or a ButtonLInk. I'll fix it. Thanks

src/components/layout/GradientBottomActions.tsx
```diff
@@ -1,18 +1,34 @@
...
+
 export type GradientBottomActions = WithTestID<{
   transitionAnimStyle: Animated.AnimateStyle<StyleProp<ViewStyle>>;
   dimensions: GradientBottomActionsDimensions;
   // Accepted components: ButtonSolid, ButtonLink
   // Don't use any components other than this, please.
-  primaryAction?: React.ReactNode;
-  secondaryAction?: React.ReactNode;
+  primaryAction?: ButtonSolidAction;
+  secondaryAction?: ButtonLinkAction;
```
dmnplb, line 31: A definite improvement over the first iteration. What do you think about assigning the `Button[...]Props` directly? Does it make sense?

```suggestion
  primaryAction?: ButtonSolidProps;
  secondaryAction?: ButtonLinkProps;
```

src/components/layout/GradientBottomActions.tsx
```diff
@@ -1,18 +1,23 @@
...
 
 export type GradientBottomActions = WithTestID<{
   transitionAnimStyle: Animated.AnimateStyle<StyleProp<ViewStyle>>;
   dimensions: GradientBottomActionsDimensions;
-  // Accepted components: ButtonSolid, ButtonLink
-  // Don't use any components other than this, please.
-  primaryAction?: React.ReactNode;
-  secondaryAction?: React.ReactNode;
+  // Accepted components: ButtonSolid for the primaryAction, ButtonLink for the secondaryAction
+  primaryActionProps?: ButtonSolidProps;
```
dmnplb, line 19: The primary button should be always rendered with `fullWidth` set to `true`, so we can exclude that prop to avoid redundant (and ineffective) props:
```suggestion
  primaryActionProps?: Omit<ButtonSolidProps, "fullWidth">;
```

src/components/layout/GradientBottomActions.tsx
```diff
@@ -99,7 +104,7 @@ export const GradientBottomActions = ({
       ]}
     />
     <View style={styles.buttonContainer} pointerEvents="box-none">
-      {primaryAction}
+      {primaryAction && <ButtonSolid {...primaryAction}></ButtonSolid>}
```
dmnplb, line 107: In this specific component, `ButtonSolid` should always be rendered with the `fullWidth` variant, then we can include it directly:
```suggestion
      {primaryAction && <ButtonSolid fullWidth {...primaryAction}></ButtonSolid>}
```

dmnplb (changes requested):
Just final changes and then we can get it approved 👍🏻

dmnplb (approved):

PR #127, opened 2023-11-02
[IOPLT-133] Implements NumberPad component

CrisTofani:
## Short description
This PR creates the new NumberPad component

- Implements new `NumberPad` component

## How to test
Check related screen in example app

https://github.com/pagopa/io-app-design-system/assets/3959405/2a69bdb6-1158-49ab-a3c0-bbd39f907e85


src/components/numberpad/NumberPad.tsx
```diff
@@ -0,0 +1,135 @@
...
+import React, { ComponentProps, useRef } from "react";
+import { View } from "react-native";
+import { BiometricsValidType } from "../../utils/types";
+import { IONumberPadButtonStyles, IOStyles } from "../../core";
+import { VSpacer } from "../spacer";
+import { IconButton } from "../buttons";
+import { IOIconSizeScale, IOIcons } from "../icons";
+import { NumberButton } from "./NumberButton";
+
+type BiometricAuthProps =
```
drmarro, line 11: can this type be just :
```
type BiometricAuthProps = {
  biometricType?: BiometricsValidType;
  onBiometricPress?: () => void;
  biometricAccessibilityLabel?: string;
};
```

src/components/numberpad/NumberPad.tsx
```diff
@@ -0,0 +1,135 @@
...
+  };
+
+  const onDeletePress = () => {
+    numberPadValue.current = numberPadValue.current.slice(0, -1);
+    onValueChange(numberPadValue.current);
+  };
+
+  type ButtonType = "biometric" | "delete";
+
+  const RowButtons = ({
```
drmarro, line 71: Could this component be extracted as a standalone, reusable component, separate from the NumberPad?

CrisTofani: Actually this should be only a local component avoiding code duplication to repeat button lines. Do you see any other possible usage of the component? 😄

src/components/numberpad/NumberPad.tsx
CrisTofani: Using the union type here defined we can handle the constraint of having all three props none of them defined at the same time: if we have a biometric option the tercet must be valorized correctly

dmnplb:
As I can see from the simulator (and now from the attached video), the component does not seem to respect the experimental design mode.

dmnplb (changes requested):
As I can see from the simulator (and now from the attached video), the component does not seem to respect the experimental design mode.

CrisTofani:
> As I can see from the simulator (and now from the attached video), the component does not seem to respect the experimental design mode (in the disabled state)

fixed in [d684d67](https://github.com/pagopa/io-app-design-system/pull/127/commits/d684d678848fc1a8f8350871aad76d188e0b263f)

dmnplb (approved):

PR #130, opened 2023-11-08
fix: HeaderSecondLevel background color don't covers safearea

CrisTofani:
## Short description
Fixes the header background color when transparent mode is enabled

- Feature A

## How to test
Check the HeaderSecondLevel page on example app add a background color to any element, header should cover it.


mastro993 (approved):
I've tested the header on the Example app and works pretty well!

PR #133, opened 2023-11-14
[IOPLT-239] Creates the `Stepper` component

CrisTofani:
## Short description
This PR creates the `Stepper` component

- Creates Stepper component
- Include standard example in example App
- Include header integration in example App

## How to test
Check Example app for usage and UI Checking


src/components/stepper/Stepper.tsx
```diff
@@ -0,0 +1,73 @@
...
+  IOSpacer,
+  IOStyles,
+  IOVisualCostants,
+  useIOExperimentalDesign
+} from "../../core";
+
+type StepperProps = {
+  steps: number;
+  currentStep: number;
+  padded?: boolean;
```
dmnplb, line 16:
```suggestion
```
We need to use this component in a fixed position, so we probably always need to add this property. We can take it for granted.

dmnplb, line 16:
```suggestion
```
We need to use this component in a fixed position, so we probably always need to add this property. We can take it for granted.

src/components/stepper/Stepper.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+const legacyColorMap: Record<string, IOColors> = {
+  active: "blue",
+  inactive: "bluegrey"
+};
+
+export const Stepper = ({
+  steps,
+  currentStep,
+  padded = false
```
dmnplb, line 33:
```suggestion
```

src/components/stepper/Stepper.tsx
```diff
@@ -0,0 +1,73 @@
...
+  );
+  const onLayout = (e: LayoutChangeEvent) => {
+    setStepWidth(e.nativeEvent.layout.width / steps);
+  };
+
+  return (
+    <View
+      style={
+        padded ? { paddingHorizontal: IOVisualCostants.appMarginDefault } : {}
+      }
```
dmnplb, line 52:
```suggestion
      style={{ paddingHorizontal: IOVisualCostants.appMarginDefault }}
```

dmnplb (changes requested):

dmnplb (approved):

thisisjp:
@CrisTofani @dmnplb looks like the unactive step has a darker gray compared to the design 👀

dmnplb:
@CrisTofani @thisisjp That's correct, but I missed it completely because the new version (with experimental DS on) has the correct chromatic value. Let's fix it before the next release.

PR #135, opened 2023-11-15
[IOPLT-235] Create `ListItemHeader` component

CrisTofani:
## Short description
Implements new `ListItemHeader` component

- new `ListItemHeader` component

## How to test
Check List items page in example app


src/components/listitems/ListItemHeader.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+type EndElementProps =
+  | ButtonLinkActionProps
+  | IconButtonActionProps
+  | BadgeProps;
+
+export type ListItemHeader = WithTestID<{
+  label: string;
+  numberOfLines?: number;
+  icon?: IOIcons;
```
dmnplb, line 38: `ListItemHeader` will be used in several screens. So we should be more open about custom visual attributes. In most cases, the `icon` prop with a default color will probably be enough. But in some specific cases, an `iconProps` object (like `componentProps`) might fit better. What do you think?

src/components/listitems/ListItemHeader.tsx
```diff
@@ -0,0 +1,138 @@
...
+      testID={testID}
+      accessible={endElement === undefined ? true : false}
+      accessibilityLabel={listItemAccessibilityLabel}
+    >
+      <View style={IOListItemStyles.listItemInner}>
+        {icon && (
+          <View style={{ marginRight: IOListItemVisualParams.actionMargin }}>
+            <Icon
+              name={icon}
+              color="grey-450"
```
dmnplb, line 122:
```suggestion
              color={theme["icon-decorative"]}
```
What about adding a reference to the `icon-decorative` key from the `IOTheme` object? (I know the other list items should be updated, but let's start with this one)

src/components/listitems/ListItemHeader.tsx
CrisTofani: if we keep the color fixed to `theme["icon-decorative"]` the only props accepted would be icon name and size... so you think it would be useful?

dmnplb: My proposal is not to keep the color fixed (and not configurable), but to let the developer choose the icon color **if needed** (let's think about the specific pagoPA icon color we've already seen). I wouldn't make the size configurable. Of course `theme["icon-decorative"]` would be the correct fallback color if not specified.

src/components/listitems/ListItemHeader.tsx
```diff
@@ -0,0 +1,146 @@
...
+      testID={testID}
+      accessible={endElement === undefined ? true : false}
+      accessibilityLabel={listItemAccessibilityLabel}
+    >
+      <View style={IOListItemStyles.listItemInner}>
+        {iconName && (
+          <View style={{ marginRight: IOListItemVisualParams.actionMargin }}>
+            <Icon
+              name={iconName}
+              color={iconColor ? iconColor : theme["icon-decorative"]}
```
CrisTofani, line 130:
```suggestion
              color={iconColor ?? theme["icon-decorative"]}
```

dmnplb (approved):

PR #139, opened 2023-11-17
[IOBP-403] `ListItemRadioWithAmount` addition

forrest57:
## Short description
added said component

- added component
- added required icon

## How to test
Describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.


src/components/icons/Icon.tsx
```diff
@@ -379,7 +379,8 @@ export const IOIcons = {
   systemSettingsiOS: IconSystemSettingsiOS,
   systemToggleInstructions: IconSystemToggleInstructions,
   systemAppsAndroid: IconSystemAppsAndroid,
-  systemNotificationsInstructions: IconSystemNotificationsInstructions
+  systemNotificationsInstructions: IconSystemNotificationsInstructions,
+  doubleSarsFilled: IconStarFilled
```
dmnplb, line 383:
```suggestion
  doubleStarsFilled: IconDoubleStarsFilled
```
I don't know if you checked out the example app, but `IconStarFilled` is not the right component, according to the new component added in the `svg` folder.

Also consider moving the key up, because the last keys are used for other sets (as you can see from the `system` prefix).

src/components/icons/svg/IconDoubleStarsFilled.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from "react";
+import { Path, Svg } from "react-native-svg";
+import { SVGIconProps } from "../Icon";
+const IconDoubleStarsFilled = ({ size, style, ...props }: SVGIconProps) => (
+  <Svg width={size} height={size} viewBox="0 0 16 17" style={style} {...props}>
+    <Path
+      fill="#5517E3"
```
dmnplb, line 7:
```suggestion
      fill="currentColor"
```
Better to leave the icon color-agnostic and set the color using the general color palette of the app (`IOColors`)

src/components/listitems/ListItemTransactionRadio.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const pressHandler = () => {
+    RNReactNativeHapticFeedback.trigger("impactLight");
+    setToggleValue(val => !val);
+  };
+  const isExperimental = useIOExperimentalDesign();
+  return (
+    <PressableListItemBase onPress={pressHandler}>
+      <View>
+        <LabelSmall color="black" weight="SemiBold">
+          Label
```
dmnplb, line 37:
```suggestion
          Label
```
Hardcoded label? 🤔

dmnplb (changes requested):
You should also run the `generateNewTimestamp` script (in the `icons` folder) to avoid overwriting the `IconDoubleStarsFilled` component when the `generateNewIcons` script is run again.

src/components/listitems/ListItemTransactionRadio.tsx
```diff
@@ -0,0 +1,61 @@
...
+import RNReactNativeHapticFeedback from "react-native-haptic-feedback";
+import { IOStyles, useIOExperimentalDesign } from "src/core";
+import { Icon } from "../icons";
+import { AnimatedRadio } from "../radio/AnimatedRadio";
+import { HSpacer, VSpacer } from "../spacer";
+import { H6 } from "../typography";
+import { LabelSmall } from "../typography/LabelSmall";
+import { PressableListItemBase } from "./PressableListItemsBase";
+
+type ListItemTransactionRadioProps = {
```
dmnplb, line 12: We could try to find a new name, considering that this is a specific instance of a `ListItemRadio` with a clear purpose. We already know that `amountString` is mandatory, so a more specific name could be added. For example, `ListItemRadioChooseAmount` or something similar.

forrest57: To be honest, `amountString` to me sounds like the correct name since it **is** an amount -- just not a straight amount, because we don't do any formatting, and it sounds specific enough to me.
Maybe `formattedAmountString` since it is an amount string, which has to be formatted prior to using it; what do you think?

dmnplb: Mine was not a comment on the prop name, which is perfectly fine (I like your proposal, too). I was referring to the name of the entire component. Maybe I should have pointed the comment to the correct line.

forrest57: oh welp, my bad 😅 
Since it is marked as `ListItemTransaction` in figma, I thought it would make sense to name it like that, especially since it's strongly bound to transactions; also I think "chooseAmount" doesn't really do it justice, since there is no "multiple choice" in the component itself; what do you think?

src/components/listitems/ListItemTransactionRadio.tsx
forrest57: whoopsie, that's a hard miss 😅 
addressed in [86ab979](https://github.com/pagopa/io-app-design-system/pull/139/commits/86ab9790177fe4ed7ef2c9560635c01f03991d49)

src/components/icons/svg/IconDoubleStarsFilled.tsx
forrest57: addressed in [86ab979](https://github.com/pagopa/io-app-design-system/pull/139/commits/86ab9790177fe4ed7ef2c9560635c01f03991d49)

src/components/listitems/ListItemTransactionRadio.tsx
dmnplb: > also I think "chooseAmount" doesn't really do it justice, since there is no "multiple choice" in the component itself; what do you think?

It's a good observation, but a single `RadioItem` doesn't make sense. If there are at least two `RadioItem` components, with `amountString` as a mandatory prop, we have a `ChooseAmount` scenario. That's why I suggested that name.

My main concern is that `ListItemTransactionRadio` doesn't explain its purpose very well. How about `ListItemRadioTransactionWithAmount` or `ListItemRadioWithAmount`? I know it's pretty wordy, but better to be clear than cryptic.

(It is better to use `ListItemRadio...` than `ListItemTransaction...`, because with the former we have automatically grouped the different `ListItemRadio` components)

forrest57: addressed in [31959ac](https://github.com/pagopa/io-app-design-system/pull/139/commits/31959aca9ea86cd855145e281a7776ed49167f30), LMK what you think about the `ListItemRadioWithAmount` name 👍

dmnplb (approved):
Approved with these last minute changes:
- Rename the icon to `sparkles`
- Export the `ListItemRadioWithAmountProps` type
- Update test name and associated test snapshots

Thank you for the contribution ✌🏻

PR #143, opened 2023-11-22
[IOPLT-171] Introduces `CodeInput` component

CrisTofani:
## Short description
This PR introduces `CodeInput` component

## How to test
Check NumberPad screen to view new `CodeInput` component features

https://github.com/pagopa/io-app-design-system/assets/3959405/ae9893d9-1629-4f1f-b0cf-5806dd74a9e3


dmnplb (changes requested):
Slightly improved the shake animation, but the main issue is that `CodeInput` render the filled circle with `black` fill color, instead of the `white` one, when the screen background is blue

CrisTofani:
> Slightly improved the shake animation, but the main issue is that `CodeInput` renders the filled circle with `black` fill color, instead of the `white` one if the screen background is blue

Not a real bug, just didn't map the correct variant to the `blueBackground` state, fixed in [b2350ec](https://github.com/pagopa/io-app-design-system/pull/143/commits/b2350ec32d6b0678c132253ddf2d70cebf5e398a)

dmnplb (approved):

PR #151, opened 2023-11-27
adds some storybook stories

StefanoFrontini:
## Short description
adds some storybook stories: spinner, logos, checkbox, avatar, banner and alert 

New stories:

- Alert
-Banner,
- Checkbox
     - AnimatedCheckbox, 
     - AnimatedMessageCheckbox, 
     -CheckboxLabel)
- LoadingSpinner
- LogoPayment 
   - LogoPaymentCard 
   - LogoPaymentExt
- Avatar

## How to test
Launch storybook with yarn storybook


stories/alert/Alert.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+      return (
+        <Alert
+          content={content}
+          title={title}
+          viewRef={viewRef}
+          variant={variant}
+        />
+      );
+    };
+    return AlertWithRef();
```
CrisTofani, line 44:
```suggestion
    const viewRef = React.createRef<View>();
    return (
      <Alert
        content={content}
        title={title}
        viewRef={viewRef}
        variant={variant}
      />
    );
```

I think we can omit to define and call a new function letting render function return directly the component, what do you think?

CrisTofani (changes requested):

stories/alert/Alert.stories.tsx
StefanoFrontini: I agree, in this case it can be simplified.  However this setup is needed if instead of React.createRef we use React hooks like useRef.

CrisTofani: yes I know that it is needed for the usage of ref, my suggestion is to replace the function definition + directly calling the function by just returning the component in the render argument of the story's meta.

Just like it has been done [here](https://github.com/pagopa/io-app-design-system/blob/main/stories/numberpad/NumberPad.stories.tsx#L16)

CrisTofani (approved):

PR #154, opened 2023-11-28
[IOPLT-226] Migrates `ForceScrollDownViewPage` from main app codebase

CrisTofani:
## Short description
This PR includes the Scroll to bottom component from main app code

- `ForceScrollDownView` component

## How to test
Check the related screen in example app


src/components/layout/ForceScrollDownView.tsx
```diff
@@ -0,0 +1,210 @@
...
+    </>
+  );
+};
+
+const styles = StyleSheet.create({
+  scrollDownButton: {
+    position: "absolute",
+    zIndex: 10,
+    right: 20,
+    bottom: 50
```
dmnplb, line 206: According to the design guidelines, the following values should be used instead of the `20, 50` values:
```suggestion
    right: 24,
    bottom: 24
```
Even better if we:
- Define a variable to make the configuration easier (favorite option)
- Define a couple of configurable values `{ xOffset, yOffset }` for edge cases (not entirely convinced about it)

dmnplb (changes requested):
Also, have you tried using `ForceScrollDownViewPage` with other custom `ScrollView` components? Or are they mutually exclusive?

src/components/layout/ForceScrollDownView.tsx
CrisTofani: Addressed in [0946240](https://github.com/pagopa/io-app-design-system/pull/154/commits/0946240b39d40fb78af563dfa13efaa308dbc981)

dmnplb (approved):

PR #157, opened 2023-11-30
[IOBP-429] listItemHeader value variant

forrest57:
## Short description
introduction of said variant
Screenshot 2023-11-30 at 10 34 33

- added props and component for said variant
- added playground entry

## How to test
yarn test, also you can just run the example app


dmnplb (commented):
Considering that this `ListItemHeader` is a very specific component for payments, and it's not always used as a header section (see [here](https://www.figma.com/file/SDaOSTsQnJUxmoPQEfz3bJ/Ricevuta-di-pagamento-pagoPA?type=design&node-id=520-10951&mode=design&t=ZbMNszL2NX1Os89C-4) and [here](https://www.figma.com/file/KY9BpP4LaSSd18fuYMwxYK/Paga-un-avviso-pagoPA?type=design&node-id=874-17522&mode=design&t=TnNJiVRI8wDRB3oZ-4)), I think a separate component with a specific purpose is a better fit (`ListItemInfoAmount` or `ListItemAmount`?).

We can assemble it in a modular way, thanks to the typographic styles already available (H6, etc...). Also note that the new ListItem should have a loading state for the amount (see the comment [here](https://www.figma.com/file/4Pbt3wd9WoPnAXVMkAA6UT/App-IO-2.0---Design-System?type=design&node-id=8373%3A41774&mode=design&t=ozwYSFOfmIX7uZ3G-1)).

What do you think?

forrest57:
> Considering that this `ListItemHeader` is a very specific component for payments, and it's not always used as a header section (see [here](https://www.figma.com/file/SDaOSTsQnJUxmoPQEfz3bJ/Ricevuta-di-pagamento-pagoPA?type=design&node-id=520-10951&mode=design&t=ZbMNszL2NX1Os89C-4) and [here](https://www.figma.com/file/KY9BpP4LaSSd18fuYMwxYK/Paga-un-avviso-pagoPA?type=design&node-id=874-17522&mode=design&t=TnNJiVRI8wDRB3oZ-4)), I think a separate component with a specific purpose is a better fit (`ListItemInfoAmount` or `ListItemAmount`?).
> 
> We can assemble it in a modular way, thanks to the typographic styles already available (`H6`, etc...). Also note that the new ListItem should have a loading state for the amount (see the comment [here](https://www.figma.com/file/4Pbt3wd9WoPnAXVMkAA6UT/App-IO-2.0---Design-System?type=design&node-id=8373%3A41774&mode=design&t=ozwYSFOfmIX7uZ3G-1)).
> 
> What do you think?

you're right, it probably is better to divide them!
done in [2fc4a08](https://github.com/pagopa/io-app-design-system/pull/157/commits/2fc4a0847b1ced1c9985cb9a633219c60554ae88), LMK what you think and sorry for making you wait 😅

dmnplb (approved):

PR #160, opened 2023-12-04
chore: [IOBP-450] Add onPress event to `ListItemRadioWithAmount`

Hantex9:
## Short description
This PR adds the `onPress` event to the `ListItemRadioWithAmount` to handle the press event.

- Added `onPress` event to the component property
- Added a custom hook to set the disabled/enabled flag based on the `selected` property


src/components/listitems/ListItemRadioWithAmount.tsx
```diff
@@ -25,16 +26,21 @@ export type ListItemRadioWithAmountProps = {
...
+    onPress?.(!toggleValue);
   };
   const isExperimental = useIOExperimentalDesign();
 
   const interactiveColor: IOColors = isExperimental ? "blueIO-500" : "blue";
   const suggestColor: IOColors = "hanPurple-500";
 
+  React.useEffect(() => {
+    setToggleValue(props.selected ?? false);
+  }, [props.selected]);
```
CrisTofani, line 42: I would avoid using an effect to handle the state change, since we have a `pressHandler` let's trigger this state change inside that handler. What do you think?

CrisTofani (changes requested):

src/components/listitems/ListItemRadioWithAmount.tsx
Hantex9: Look at [this PR](https://github.com/pagopa/io-app/pull/5302), as you can see, the user can choose only one element from the list, so if the user selects an item, the other items must be deselected. We cannot achieve this behavior if we don't handle the state change through an effect since the `toggleValue` must be handled to his parent as you can see [in this file](https://github.com/pagopa/io-app/pull/5302/files#diff-35eca3fdb51b2ba91abc3a60e450c2d7a39e7aa6d482fa1373f1a74e51479ba9) at row `185`. What do you think about it?

CrisTofani: you can always change the state in the `pressHandler` function at line 31 avoiding the usage of an effect the same behaviour has been implemented in `ListItemRadio` without any effect handling https://github.com/pagopa/io-app-design-system/blob/276bd17760cb929960bba11fed7de61962a680b1/src/components/listitems/ListItemRadio.tsx#L137

Hantex9: Doing in this way we cannot achieve the possibility of having only one element selected, removing the effect and adding the suggested implementation `setToggleValue` inside the `pressHandler` function, this is the behavior:

|With effect|Without effect|
|-|-|
| <video src="https://github.com/pagopa/io-app-design-system/assets/34343582/bcbd98c8-2e2a-4160-99e1-772571e5780f" />|<video src="https://github.com/pagopa/io-app-design-system/assets/34343582/40f24c4f-1b58-4d51-a200-7b36a0f14b9c" />

CrisTofani: For this kind of usage we're extending the `RadioGroup` in order to let the developer use the `ListItemRadioWithAmount` or other kind of radio components that may be implemented in future.

This is the related story on our Jira board https://pagopa.atlassian.net/browse/IOPLT-296

CrisTofani (commented):
@Hantex9 PR #166 includes the changes we were discussing in the comments before. 

It actually adds the possibility to handle `ListItemRadioWithAmount` in a `RadioGroup`, and adds the onChange prop. 

Do yout think we can close this PR?

Hantex9:
> @Hantex9 PR #166 includes the changes we were discussing in the comments before.
> 
> It actually adds the possibility to handle `ListItemRadioWithAmount` in a `RadioGroup`, and adds the onChange prop.
> 
> Do yout think we can close this PR?

Yes, absolutely! Thank you!

CrisTofani:
closing since https://github.com/pagopa/io-app-design-system/pull/166 solves the same issue

PR #161, opened 2023-12-05
[IOPLT-236] Adds the new `OTPInput` component

CrisTofani:
## Short description
This PR implements the new `OTPInput` component.

- `BoxedInput` component for number/secret representation
- `OTPInput` component to mask a real hidden TextInput

## How to test
Check the example app related page

https://github.com/pagopa/io-app-design-system/assets/3959405/75b3fc97-c242-4080-ae27-ffde37008aca

### DIFF
--- a/src/components/icons/Icon.tsx
+++ b/src/components/icons/Icon.tsx
@@ -5,10 +5,12 @@ import { IOColors } from "../../core/IOColors";
 /* Icons */
 import IconAbacus from "./svg/IconAbacus";
 import IconAdd from "./svg/IconAdd";
+import IconAddSmall from "./svg/IconAddSmall";
 import IconAgreement from "./svg/IconAgreement";
 import IconAmount from "./svg/IconAmount";
 import IconAnalytics from "./svg/IconAnalytics";
 import IconArchive from "./svg/IconArchive";
+import IconArchiveFilled from "./svg/IconArchiveFilled";
 import IconArrowBottom from "./svg/IconArrowBottom";
 import IconArrowLeft from "./svg/IconArrowLeft";
 import IconArrowRight from "./svg/IconArrowRight";
@@ -37,6 +39,7 @@ import IconCategSustainability from "./svg/IconCategSustainability";
 import IconCategTelco from "./svg/IconCategTelco";
 import IconCategTravel from "./svg/IconCategTravel";
 import IconCategWellness from "./svg/IconCategWellness";
+import IconChange from "./svg/IconChange";
 import IconChat from "./svg/IconChat";
 import IconCheckTick from "./svg/IconCheckTick";
 import IconCheckTickBig from "./svg/IconCheckTickBig";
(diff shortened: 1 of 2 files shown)


StefanoFrontini (commented):
When clicking on an input box and then pressing a keybord key  nothing happens (only the , and . seems to work)

CrisTofani:
> When clicking on an input box and then pressing a keybord key nothing happens (only the , and . seems to work)

addressed in https://github.com/pagopa/io-app-design-system/pull/161/commits/440757f44c016b8841414693e015ed91e27b5d84

src/components/otpInput/OTPInput.tsx
```diff
@@ -0,0 +1,130 @@
...
+        const timer = setTimeout(() => {
+          setHasError(false);
+          setInputValue("");
+          onValueChange("");
+        }, 500);
+        return () => clearTimeout(timer);
+      }
+    }
+    return;
+  }, [value, onValidate, length, onValueChange, translate, shakeAnimation]);
```
StefanoFrontini, line 79:
```suggestion
const inputRef = useRef<TextInput>(null);

  const timerRef = useRef<NodeJS.Timeout>();
  const handleValidate = (val: string) => {
    if (!onValidate || val.length < length) {
      return;
    }

    const isValid = onValidate(val);

    if (!isValid) {
      setHasError(true);
      triggerHaptic("notificationError");

      // eslint-disable-next-line functional/immutable-data
      translate.value = shakeAnimation();

      if (timerRef.current) {
        clearTimeout(timerRef.current);
      }

      // eslint-disable-next-line functional/immutable-data
      timerRef.current = setTimeout(() => {
        setHasError(false);
        setInputValue("");
        onValueChange("");
      }, 500);
    }
  };
  const handleChange = (value: string) => {
    if (value.length > length) {
      return;
    }
    setInputValue(value);
    onValueChange(value);
    handleValidate(value);
  };

`

File: src/components/otpInput/OTPInput.tsx [hunk 2]
```diff
@@ -0,0 +1,130 @@
+import * as React from "react";
+import { useEffect, useState } from "react";
```
StefanoFrontini, line 2:
```suggestion
import { useRef, useState } from "react";
```

src/components/otpInput/OTPInput.tsx
CrisTofani: I'm just missing one point on this suggestion, what would trigger the validation callback?

StefanoFrontini: It should be triggered by the handleChange function

CrisTofani: Tested out, it actually works perfectly!

Just some changes:
- I left the `createRef` for `inputRef` cause of type specifications
- I left the `useEffect` to handle the timeout clear when the component is unmounted

CrisTofani: [513c46a](https://github.com/pagopa/io-app-design-system/pull/161/commits/513c46afa9a0aff1ba79b7afdffaab70fa2b6946)

StefanoFrontini (approved):

dmnplb (approved):

PR #164, opened 2023-12-12
[IOPLT-298] - accordion icon

StefanoFrontini:
## Short description
adds a possible icon to the top of the header that opens the accordion

- The accordion item now has support for a possible icon in  the header title.  

## How to test
Open example app - accordion


src/components/accordion/AccordionItem.tsx
```diff
@@ -34,6 +35,9 @@ const accordionIconMargin: IOSpacingScale = 8;
 const accordionBorder: IOColors = "grey-200";
 const accordionBackground: IOColors = "white";
 
+// Icon size
+const iconSize: IOIconSizeScale = 20;
```
dmnplb, line 39: According to the design guidelines:
```suggestion
const iconSize: IOIconSizeScale = 24;
```

src/components/accordion/AccordionItem.tsx
```diff
@@ -97,8 +101,23 @@ export const AccordionItem = ({ title, body }: AccordionItem) => {
...
       >
         <View style={styles.textContainer}>
-          <View style={{ flexShrink: 1, marginRight: accordionIconMargin }}>
-            <H6 color="black">{title}</H6>
+          <View
+            style={[
+              IOStyles.row,
+              IOStyles.alignCenter,
+              {
+                flexShrink: 1
```
dmnplb, line 109: The const `accordionIconMargin` has been added to add a margin between the text and the chevron icon. With this rewrite, we should define a new `accordionChevronMargin`: 
```suggestion
                flexShrink: 1,
                marginRight: accordionChevronMargin // to be added
```
The original `accordionIconMargin` should be `12` (not `8`), while the new `accordionChevronMargin` should have the original value, so `8`

dmnplb (changes requested):

CrisTofani (approved):

PR #168, opened 2023-12-13
[IOPLT-181] - snapshots experimental

StefanoFrontini:
## Short description
Adds snapshots tests for the new design system variant.

Now snapshots tests cover also the new variant as well as legacy variant

## How to test
yarn test


src/core/IODSExperimentalContextProvider.tsx
```diff
@@ -37,3 +37,19 @@ export const IODSExperimentalContextProvider = ({
...
+  );
+
+  return (
+    <IOExperimentalDesignContext.Provider
+      value={{ isExperimental, setExperimental }}
+    >
+      {children}
+    </IOExperimentalDesignContext.Provider>
+  );
+};
```
CrisTofani, line 55: instead of creating a new instance of our context provider what about using the original one setting to `true` the `isExperimentaEnabled` prop ?

CrisTofani:
```suggestion
```

src/utils/testing.tsx
```diff
@@ -0,0 +1,14 @@
...
+export const TestRendererWithExperimentalEnabledContextProvider = (
+  ui: React.ReactElement,
+  options?: TestRenderer.TestRendererOptions
+) =>
+  TestRenderer.create(
+    <IODSExperimentalEnabledContextProvider>
+      {ui}
+    </IODSExperimentalEnabledContextProvider>,
+    options
+  );
```
CrisTofani, line 14: ditto
```suggestion
import { IODSExperimentalContextProvider } from "../core/IODSExperimentalContextProvider";

export const TestRendererWithExperimentalEnabledContextProvider = (
  ui: React.ReactElement,
  options?: TestRenderer.TestRendererOptions
) =>
  TestRenderer.create(
    < IODSExperimentalContextProvider isExperimentaEnabled={true}>
      {ui}
    </IODSExperimentalContextProvider>,
    options
  );
```

CrisTofani (changes requested):

src/core/IODSExperimentalContextProvider.tsx
StefanoFrontini: You are absolutely right! Thanks!

CrisTofani (approved):

PR #185, opened 2024-02-05
chore: [IOBP-530] Edited `ModuleCheckout` component without CTA

Hantex9:
## Short description
This PR introduces the possibility of having the `ModuleCheckout` component without a CTA, using it only as a presentational component.

- Edited the `ctaText` as an optional prop;
- When the `ctaText` is not defined, it renders the component only in presentation mode without the pressable action and the CTA label

## Preview

https://github.com/pagopa/io-app-design-system/assets/34343582/59fb1075-6687-46ae-adbe-b8c8d8451694


src/components/modules/ModuleCheckout.tsx
```diff
@@ -43,29 +43,37 @@ export const ModuleCheckout = (props: ModuleCheckoutProps) => {
...
   );
+
+  if (props.ctaText) {
+    return (
+      <PressableModuleBase onPress={props.onPress}>
+        {/*
+          we don't want to let the `space-between`
+          handle spacing for the logo/text section,
+          so we use a row and a marginEnd on the logo
+        */}
```
forrest57, line 69: I think this now belongs to the `ModuleBaseContent`, what do you think?

Hantex9: Addressed into 56ad7deff683e86c9f63e0470dc2530f9b182b7a

mastro993 (changes requested):

src/components/modules/ModuleCheckout.tsx
```diff
@@ -43,29 +43,37 @@ export const ModuleCheckout = (props: ModuleCheckoutProps) => {
 
   const paymentLogoEndMargin: IOSpacingScale = 12;
 
-  return (
-    <PressableModuleBase onPress={props.onPress}>
+  const ModuleBaseContent = (
```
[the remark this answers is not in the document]
Hantex9: Since it is a local component, what do you think about the implementation in this commit (fb1dc45b64cb85ef93554b1baad787035306f647)? Doing in this way, when the `ModuleBaseContent` is invoked, it will be a simple tag `<ModuleBaseContent />`

NB: In the commit fb1dc45b64cb85ef93554b1baad787035306f647 has been fixed also the alignment of the title and CTA button if present a title is very log

mastro993 (approved):
LGTM

PR #188, opened 2024-02-06
[chore] Add ModuleAttachment, fix accessibility on ListItemHeader

Vangaorth:
## Short description
This PR add the ModuleAttachment component and fixes accessibility on Android for ListItemHeader

- Added ModuleAttachment
- ListItemHeader's label ignores accessibility on Android if the parent view is marked as accessibile (this avoid a double reading of such label - iOS is not affected)

### DIFF
--- a/src/components/listitems/ListItemHeader.tsx
+++ b/src/components/listitems/ListItemHeader.tsx
@@ -125,7 +125,10 @@ export const ListItemHeader = ({
       accessible={endElement === undefined ? true : false}
       accessibilityLabel={listItemAccessibilityLabel}
     >
-      <View style={IOListItemStyles.listItemInner}>
+      <View
+        style={IOListItemStyles.listItemInner}
+        importantForAccessibility={endElement ? "auto" : "no-hide-descendants"}
+      >
         {iconName && (
           <View style={{ marginRight: iconMargin }}>
             <Icon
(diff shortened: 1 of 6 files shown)


src/components/listitems/ListItemHeader.tsx
```diff
@@ -125,7 +125,10 @@ export const ListItemHeader = ({
       accessible={endElement === undefined ? true : false}
       accessibilityLabel={listItemAccessibilityLabel}
     >
-      <View style={IOListItemStyles.listItemInner}>
+      <View
+        style={IOListItemStyles.listItemInner}
+        importantForAccessibility={endElement ? "auto" : "no-hide-descendants"}
```
CrisTofani, line 130: maybe we should handle `elementsHiddenForAccessibility` too in order to have the same behaviour both in iOS and Android, what do you think?

CrisTofani (changes requested):
Apart from required changes, we need to add the new component in the example app and storybook for documentation purpose

src/components/listitems/ListItemHeader.tsx
CrisTofani: Perfect! 👌🏻

Vangaorth:
> Apart from required changes, we need to add the new component in the example app and storybook for documentation purpose

Done it!

CrisTofani (approved):

PR #190, opened 2024-02-09
[chore] Icon accessibility

Vangaorth:
This PR depends on #188 

## Short description
This PR fixes the accessibility on the Icon component, which was always disabled.

- specific iOS and Android properties are toggled based on the accessibility data

## How to test
Check that both TalkBack and VoiceOver read the accessibility label when it is specified and ignore it when it is not.


CrisTofani (approved):
lgtm

PR #195, opened 2024-02-19
docs: [IOAPPX-251] Update README.md

LazyAfternoons:
## Short description
This PR updates the `README.md` by including instructions about wrapping the app with `SafeAreaProvider` from `react-native-safe-area-context`. This is needed because a lot of components use the `useSafeAreaInsets` hook which leads to runtime errors if they are being used without the safe area context.

There's no need to install the dependency directly as it is already being declared as `peerDependency`, thus the consumer should be warned about its absence, eventually.

## How to test
Check the new README.md preview [here](https://github.com/pagopa/io-app-design-system/blob/IOAPPX-251-update-readme/README.md).


CrisTofani (commented):
Would you mind adding the same changes to `stories/markdown/GettingStarted.md` ?

LazyAfternoons:
> Would you mind adding the same changes to `stories/markdown/GettingStarted.md` ?

Done in c9518408b378d5c3e43af6e24cca39b81ae5c9f9 thanks.

dmnplb (approved):

dmnplb:
LGTM. Let's wait @CrisTofani for final approval 👍🏻

CrisTofani (approved):

PR #218, opened 2024-03-15
fix: header accessibility on empty title should avoid autofocus

CrisTofani:
## Short description
This PR prevents accessibility focus on empty header title

- HeaderSecondLevel accessibility props

## How to test
Test accessibility on real device to check TB/VO behaviour


shadowsheep1 (approved):
LGTM

PR #219, opened 2024-03-15
chore: [IOBP-585] Adjustment on the `ListItemRadioWithAmount` long label

Hantex9:
## Short description
This PR adjusts the `ListItemRadioWithAmount` when provided with a long label

- Added a flex property to the label wrapper
- Added a max number of lines to 1 in order to truncate the label if too long

## Preview
|Before|After|
|-|-|
||


example/src/pages/ListItem.tsx
```diff
@@ -683,6 +683,10 @@ const renderListItemRadioWithAmount = () => (
         label="Banca Malintesa"
         formattedAmountString={"4,50 €"}
       />
+      <ListItemRadioWithAmount
```
dmnplb, line 686:
```suggestion
      <Divider />
      <ListItemRadioWithAmount
```

src/components/listitems/ListItemRadioWithAmount.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from "react";
 import { StyleSheet, View } from "react-native";
 import RNReactNativeHapticFeedback from "react-native-haptic-feedback";
-import { IOColors, useIOTheme } from "../../core";
+import { IOColors, IOStyles, useIOTheme } from "../../core";
```
dmnplb, line 4: Remove `IOStyles` because it's not necessary (see below)
```suggestion
import { IOColors, useIOTheme } from "../../core";
```

src/components/listitems/ListItemRadioWithAmount.tsx
```diff
@@ -40,8 +40,8 @@ export const ListItemRadioWithAmount = (
 
   return (
     <PressableListItemBase onPress={pressHandler}>
-      <View>
-        <LabelSmallAlt color={theme["textBody-default"]}>
+      <View style={IOStyles.flex}>
```
dmnplb, line 43: In this case, `flexShrink` is enough. You don't need to use the nuclear option (`flex: 1`): 
```suggestion
      <View style={{ flexShrink: 1 }}>
```
It's also better to add a `HSpacer` in the right `View` (just before `H6`) to add a safe margin between the two blocks

dmnplb (changes requested):

src/components/listitems/ListItemRadioWithAmount.tsx
Hantex9: Addressed into Addressed into bc0618b324d16a9a7098c9df9195467129d5eb4a

dmnplb (approved):

PR #234, opened 2024-03-29
[IOAPPX-272] Improve accessibility by adding `accessibilityState` to `Button…` components

dmnplb:
## Short description
This PR adds the `accessibilityState` prop to all the `Button…` components

- Add `accessibilityState`

## How to test
N/A


src/components/buttons/ButtonOutline.tsx
```diff
@@ -333,6 +333,7 @@ export const ButtonOutline = React.forwardRef<View, ButtonOutline>(
         accessibilityLabel={accessibilityLabel || label}
         accessibilityHint={accessibilityHint}
         accessibilityRole={"button"}
+        accessibilityState={{ disabled: disabled || false }}
```
CrisTofani, line 336:
```suggestion
        accessibilityState={{ disabled }}
```

src/components/buttons/ButtonSolid.tsx
```diff
@@ -265,7 +265,10 @@ export const ButtonSolid = React.forwardRef<View, ButtonSolidProps>(
         // Using || operator because empty string is not an accepted value
         accessibilityLabel={accessibilityLabel || label}
         accessibilityHint={accessibilityHint}
-        accessibilityState={{ busy: loading }}
+        accessibilityState={{
+          busy: loading,
+          disabled: disabled || false
```
CrisTofani, line 270:
```suggestion
          disabled
```

CrisTofani (changes requested):
Some minor suggestion on some code smells

src/components/buttons/ButtonSolid.tsx
dmnplb: It doesn't work because we pick the `disabled` prop from Pressable and it could be also undefined. Look at the last commit where I fixed the typescript error

CrisTofani (approved):

PR #247, opened 2024-04-16
chore: add avatar prop to `ListItemNav` component

adelloste:
## Short description
This PR introduces the possibility to set an avatar in the `ListItemNav` component 

- updated `ListItemNav` component to pass avatarProps

## How to test
Launch the example app and check the ListItems section


dmnplb (approved):
Just made two changes:
- Added `withMargin` utility function to add each graphic asset to the same `View` with the same margin
- Removed the `size` prop from `Avatar` because we shouldn't accept `Avatar` in the larger variant

What do you think?

adelloste:
> Just made two changes:
> 
> * Added `withMargin` utility function to add each graphic asset to the same `View` with the same margin
> * Removed the `size` prop from `Avatar` because we shouldn't accept `Avatar` in the larger variant
> 
> What do you think?

Great job!

PR #248, opened 2024-04-19
chore: Let OTPInput reset value on demand by parent

CrisTofani:
## Short description
This PR improve the handling on OTPInput value change to help the handling of the value by the parent component.
Adds accessibilityLabel as prop to improve accessibility

- Handle the value change only by parent
- Add accessibilityLabel as component prop

## How to test
Check OTPInput component on example app page


example/src/pages/OTPInput.tsx
```diff
@@ -30,14 +31,19 @@ const OTPWrapper = ({ secret = false, validation = false }: WrapperProps) => {
...
-      value={value}
-      onValueChange={onValueChange}
-      length={OTP_LENGTH}
-      secret={secret}
-      onValidate={onValidate}
-      errorMessage={"Wrong OTP"}
-    />
+    <>
+      <OTPInput
+        value={value}
```
shadowsheep1, line 36: What about exposing also a view ref? As for a11y this'd let us focus on the element if we want. What do you think? The alternative is to wrap the componenti in an accessible `View` with a ref.

```ts
 <View accessible ref={pinPadViewRef}>
  <OTPInput
    secret
    value={pin}
    accessibilityLabel={I18n.t(
      "authentication.cie.pin.accessibility.label"
    )}
    onValueChange={setPin}
    length={CIE_PIN_LENGTH}
  />
</View>
```

shadowsheep1 (changes requested):

example/src/pages/OTPInput.tsx
CrisTofani: Added in [1a92ac5](https://github.com/pagopa/io-app-design-system/pull/248/commits/1a92ac5c175733a233550484d3ad081e534f409f)

example/src/pages/OTPInput.tsx
```diff
@@ -30,14 +31,19 @@ const OTPWrapper = ({ secret = false, validation = false }: WrapperProps) => {
...
+        accessibilityLabel={"OTP Input"}
```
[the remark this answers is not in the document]
CrisTofani: Added in [1a92ac5](https://github.com/pagopa/io-app-design-system/pull/248/commits/1a92ac5c175733a233550484d3ad081e534f409f)

src/components/otpInput/OTPInput.tsx
```diff
@@ -31,105 +33,112 @@ type Props = {
...
-                ? "focus"
-                : "default"
-            }
-            secret={secret}
-            value={inputValue[i]}
+          accessibilityLabel={accessibilityLabel}
+          accessibilityHint={accessibilityHint}
+          accessibilityValue={{ text: value }}
+        >
+          <TextInput
```
shadowsheep1, line 106: I've found that we are not exposing `InputAccessoryViewID` prop too? We actually use it in app. What do you think?

shadowsheep1 (changes requested):

src/components/otpInput/OTPInput.tsx
CrisTofani: addressed [9f8d73b](https://github.com/pagopa/io-app-design-system/pull/248/commits/9f8d73b3ac64da3c014906893c88d361f4e95925)

shadowsheep1 (approved):
LGTM

PR #252, opened 2024-05-02
chore: Expose `autoFocus` `prop` on `OTPInput`

shadowsheep1:
## Short description
This PR exposes the `autoFocus` prop on `OTPInput`.

- Add `autoFocus` on `OTPInput`.
- Add `OTPInput` with `autoFocus` on the Example App.

## How to test
Run the Example App, and inside the `OTPInput` section tap on "Show Autofocusable OTP" button to show an `OTPInput` that is focused and trigger the opening of the **Virtual Keyboard**.


src/components/otpInput/OTPInput.tsx
```diff
@@ -91,6 +93,12 @@ export const OTPInput = React.forwardRef<View, Props>(
       handleValidate(value);
     };
 
+    React.useEffect(() => {
+      if (autoFocus) {
+        setHasFocus(true);
+      }
+    }, [autoFocus]);
```
CrisTofani, line 100: Instead of working with effects, why not changing the state definition at line 55 using autoFocus value to define the hasFocus state ?

CrisTofani (changes requested):

src/components/otpInput/OTPInput.tsx
shadowsheep1: Good catch! I'll try it out.

shadowsheep1: Done! [c73a497](https://github.com/pagopa/io-app-design-system/pull/252/commits/c73a497bb1147b910949a2d334336d2281fd66fc)

CrisTofani (approved):

PR #253, opened 2024-05-02
[chore] Add optional `testID` property to `ListItemSwitch`

Vangaorth:
## Short description
This PR adds the optional property testID to ListItemSwitch.

- ListItemSwitch’s props have the testID property
- ListItemSwitch’s main view uses the provided testID property, if defined, or fallback to default one

### DIFF
--- a/src/components/listitems/ListItemSwitch.tsx
+++ b/src/components/listitems/ListItemSwitch.tsx
@@ -20,6 +20,7 @@ type PartialProps = {
   action?: SwitchAction;
   isLoading?: boolean;
   badge?: Badge;
+  testID?: string;
 };
 
 export type SwitchAction = {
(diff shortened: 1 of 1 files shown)


src/components/listitems/ListItemSwitch.tsx
```diff
@@ -20,6 +20,7 @@ type PartialProps = {
   action?: SwitchAction;
   isLoading?: boolean;
   badge?: Badge;
+  testID?: string;
```
dmnplb, line 23: Just a "cosmetic" suggestion: what about using the `withTestID` util function?

dmnplb (approved):

src/components/listitems/ListItemSwitch.tsx
Vangaorth: Good one, applied

PR #260, opened 2024-05-09
chore: [PE-594] Remove marginLeft from `ListItemNav` chevron icon

Hantex9:
## Short description
This PR removes the static marginLeft from the `ListItemNav` chevron icon

- Removed the `View` that wraps the chevron icon removing the additional marginLeft


dmnplb:
Just a note for future reference: we removed the space because the specific `chevronRightListItem` icon already has a space inside (on the left), so we don't need an additional margin.

dmnplb (approved):

PR #263, opened 2024-05-13
ForwardRef for H3

Vangaorth:
## Short description
This PR proposes a way to pass a view reference from a top screen to an inner component, using `MainScreen` and `H3` as an example.

- `H3` uses `React.forwardRef` to receive the ref as an input and it gives it to `Factory` as the second optional function parameter;
- `Factory` has a generic type that cannot be used in combination with `React.forwardRef`, that's why `ref` is received as a second optional function parameter. It is later passed down to `BaseTypography` as a standard React view's property;
- `BaseTypography` uses `React.forwardRef` to receive `ref` and to set it to its inner `Text` component.

The downside of this approach is that, when used in the main screen, the hook must not be explicitly typed, otherwise typescript complains about it, when passed to the `H3` element. E.g.:

`const aRef = useRef(null);            // This works`
`const aRef = useRef<View|null>(null); // This does not`

## How to test
Using the example app, check that the `ref.current` is properly initialized.

### DIFF
--- a/example/src/pages/MainScreen.tsx
+++ b/example/src/pages/MainScreen.tsx
@@ -9,7 +9,8 @@ import {
   ListItemNav,
   useIOExperimentalDesign,
   ListItemSwitch,
-  useIOThemeContext
+  useIOThemeContext,
+  H3
 } from "@pagopa/io-app-design-system";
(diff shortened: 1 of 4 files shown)


Vangaorth:
Replaced by #264

PR #274, opened 2024-05-28
[IOPAE-1195] Add `AvatarSearch` component

adelloste:
## Short description
This PR introduces `AvatarSearch` component, which will be used in the search results of institutions.

- Added `AvatarSearch` component


dmnplb (approved):
Just applied some small changes:
- Removed the `shape` prop from `Avatar`
- Removed the `size` prop from `AvatarSearch` (bigger sizes shouldn't be allowed)
- Removed `require()` in favour of direct import
- Changed the inner logic of `AvatarSearch` to make less verbose
- Accordingly updated snapshots and tests

PR #280, opened 2024-06-06
[IOCOM-1394,IOCOM-1478] ModulePaymentNotice and ModuleAttachment fixes

Vangaorth:
## Short description
This PR aligns the component used on ModulePaymentNotice and ModuleAttachment to the DS styles.

- ModulePaymentNotice: one-line, never-trimmed payment amount
- MouleAttachment: LabelSmallAlt instead of LabelSmall (Readex Pro instead of Titillium)

## How to test
Check that both component have the proper styles.

### DIFF
--- a/src/components/modules/ModuleAttachment.tsx
+++ b/src/components/modules/ModuleAttachment.tsx
@@ -88,7 +88,12 @@ const ModuleAttachmentContent = ({
       const activityIndicatorTestId = testID
         ? `${testID}_activityIndicator`
         : undefined;
-      return <LoadingSpinner testID={activityIndicatorTestId} />;
+      return (
+        <LoadingSpinner
+          testID={activityIndicatorTestId}
+          color={theme["interactiveElem-default"]}
+        />
+      );
     }
 
     return (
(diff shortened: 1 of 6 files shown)


src/components/modules/ModuleAttachment.tsx
```diff
@@ -103,14 +103,9 @@ const ModuleAttachmentContent = ({
   return (
     <>
       <View style={styles.innerContent}>
-        <LabelSmall
-          numberOfLines={1}
-          weight="SemiBold"
-          font="ReadexPro"
-          color="blueIO-500"
-        >
+        <LabelSmallAlt numberOfLines={1} color="blueIO-500">
```
dmnplb, line 106:
```suggestion
        <LabelSmallAlt
          numberOfLines={1}
          color={theme["interactiveElem-default"]}
        >
```
Otherwise `blueIO` will be used even when experimental DS is disabled. You should apply the same to `LoadingSpinner`, too:

```
<LoadingSpinner
  testID={activityIndicatorTestId}
  color={theme["interactiveElem-default"]}
/>
```

dmnplb (changes requested):
Also modules developed for IT Wallet flow have the same problem. I'll fix them before merge.

dmnplb (approved):

PR #284, opened 2024-06-07
[IOPID-1899] `NumberPad` optimization

ChrisMattew:
## Short description
Improved `NumberPad` component performance, avoiding unnecessary renders

- Performance improvements
- Add unitary tests for `NumberPad` component
- Add documentation in `NumberPad` and `NumberButton` components
- BREAKING CHANGE: update `NumberPad` APIs and logic**

>[!Important]
>** The pin concatenation is not handled by `NumberPad` anymore. This kind of logic will need to be handled directly where the component will be used.
## How to test
Run the `ExampleApp`, navigate in the `Components > NumberPad` section and start typing.


src/components/numberpad/NumberPad.tsx
```diff
@@ -35,79 +63,55 @@ const mapIconSpecByBiometric: Record<
...
-      IOStyles.centerJustified
-    ]}
-  >
-    {children}
-  </View>
-);
+/**
+ * This component displays a custom numeric keyboard. 
+ * 
+ * It's developed to be used in combination with the `CodeInput` component.
```
shadowsheep1, line 69: @dmnplb Is this assertion still valid? I guess that now this NumberPad is quite general purpose. What do you think?

dmnplb: @shadowsheep1 If it's decoupled, you're right: that general assumption is no longer valid and should be removed.

ChrisMattew: Alright! So, if you agree, i'll proceed removing row 69!

PR #285, opened 2024-06-10
[IOAPPX-318] Add `contrast` variant and optional `backgroundColor` prop to `HeaderSecondLevel`

dmnplb:
## Short description
This PR adds a `contrast` variant (light icons on dark background) and an optional `backgroundColor` prop to set a custom background color when there's a transition between the transparent → solid state.

- Add `variant` and `backgroundColor` props

### Preview

https://github.com/pagopa/io-app-design-system/assets/1255491/76b016f3-af08-43b4-9480-0eaf3d890c69

## How to test
1. Launch the example app
2. Check every example with `HeaderSecondLevel`, including the new custom background


hevelius (commented):
LGTM. Tested and works as expected. @mastro993 having already encountered this case, what do you think?

mastro993:
I don't know if this is expected or not (maybe is just the example), but on Android I see a white header. 

https://github.com/pagopa/io-app-design-system/assets/6160324/b6d31311-b548-47f0-a8b9-e6bfb313b414

dmnplb:
@mastro993 Considering that the blue block is just an ugly block with `position: absolute' and a negative `top' value to quickly test a dark background, it's probably limited to the example. But let me check 👍🏻

dmnplb:
@mastro993 Just fixed the example screen on both iOS and Android ✌🏻

mastro993 (approved):
LGTM! Looks good on both Android and iOS!

PR #298, opened 2024-06-25
[IOAPPX-333] Add dark mode support to all the `Module…` components

dmnplb:
> [!caution]
> This PR depends on:
> * https://github.com/pagopa/io-app-design-system/pull/296

## Short description
This PR adds the dark mode support to all the `Module…` components

- Refactor code to apply theme keys to the `color` prop in the various typographic styles
- Refactor code to use `PressableModuleBase` in all the components
- Add the new `ModuleStatic` component to render the non-interactive version of the module. It can accept:
  - Two props: `startBlock` and `endBlock`
  - Single prop: `children` (standard mode)
- Update some examples included in the `Modules` screen to test all the possible variants

### Preview

https://github.com/pagopa/io-app-design-system/assets/1255491/fe8c1a16-5b15-4605-8877-92f46a5ed549

## How to test
1. Launch the example app
2. Check the **Modules** screen, both in light and dark mode

### DIFF
--- a/src/components/modules/ModuleAttachment.tsx
+++ b/src/components/modules/ModuleAttachment.tsx
@@ -1,39 +1,20 @@
 import React, { useCallback } from "react";
-import {
-  GestureResponderEvent,
-  Pressable,
-  PressableProps,
-  StyleSheet,
-  View
-} from "react-native";
-import Animated, {
-  Extrapolate,
-  interpolate,
-  useAnimatedStyle,
-  useDerivedValue,
-  useSharedValue,
-  withSpring
-} from "react-native-reanimated";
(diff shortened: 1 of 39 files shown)


src/components/modules/ModuleAttachment.tsx
```diff
@@ -106,17 +62,17 @@ const ModuleAttachmentContent = ({
 
   return (
     <>
-      <View style={styles.innerContent}>
+      <View style={{ alignItems: "flex-start", flexShrink: 1 }}>
```
mastro993, line 65: Why did you prefer the inline styles instead of `StyleSheet` object?

mastro993 (approved):
LGTM

PR #314, opened 2024-07-24
[IOAPPX-352] Add the new IO typeface

dmnplb:
> [!note]
> This PR depends on:
> * https://github.com/pagopa/io-app-design-system/pull/312

## Short description
This PR adds the new IO typeface (temporarily named **Titillio**). This change is visible only if experimental DS is on.

- Add the new `Titillio` typeface and relative font files
- Remove references to `ReadexPro`

### Preview

https://github.com/user-attachments/assets/eb1820dc-1f82-49e9-9218-1a2430ed6477

## How to test
1. Run the example app
2. Enable experimental DS and check every screen


example/assets/css/fonts.css
```diff
@@ -1,6 +1,6 @@
 @font-face {
-    font-family: 'ReadexPro';
-    src: url('ReadexPro-Regular.ttf') format("truetype");
+    font-family: 'Titillio 3';
+    src: url('Titillio.Regular.ttf') format("truetype");
     font-weight: 400;
     font-style: normal;
 }
```
CrisTofani, line 6: shouldn't we map all the variante of Titillio?

CrisTofani (changes requested):

example/assets/css/fonts.css
dmnplb: @CrisTofani Done in the following commit → https://github.com/pagopa/io-app-design-system/pull/314/commits/ee982d13624a8fda1f29a5d94545dacf79653294
But Storybook keeps failing anyway because it can't resolve `AccessibilityInfo` from the hook `useBoldTextEnabled`

CrisTofani: testing the execution of storybook changine the `useBoldTextEnabled` as follows:

```
export const useBoldTextEnabled = () => {
  const [boldTextEnabled, setBoldTextEnabled] = useState(false);

  useEffect(() => {
    if (Platform.OS !== "web") {
      const boldTextChangedSubscription = AccessibilityInfo.addEventListener(
        "boldTextChanged",
        isBoldTextEnabled => {
          setBoldTextEnabled(isBoldTextEnabled);
        }
      );

      AccessibilityInfo.isBoldTextEnabled()
        .then(isBoldTextEnabled => {
          setBoldTextEnabled(isBoldTextEnabled);
        })
        .catch(_ => false);

      return () => {
        boldTextChangedSubscription.remove();
      };
    }
    return;
  }, []);

  return boldTextEnabled;
};
```

seems to work ok, what do you think?

dmnplb: @CrisTofani Practical solution ✌🏻 I've also pushed on the `2.x` branch by mistake, but applied here too: https://github.com/pagopa/io-app-design-system/pull/314/commits/ba846ce5eaa949fc433a5b4edba0c99aecbe0b88

CrisTofani (approved):

PR #317, opened 2024-07-29
[IOAPPX-359] Remove mandatory a11y props from some `ListItem…` components

dmnplb:
## Short description
This PR removes mandatory a11y props from some `ListItem…` components:
- `ListItemAction`
- `ListItemNavAlert`
- `ListItemInfoCopy`

## How to test
N/A

### DIFF
--- a/src/components/listitems/ListItemInfoCopy.tsx
+++ b/src/components/listitems/ListItemInfoCopy.tsx
@@ -1,4 +1,4 @@
-import React, { ComponentProps, useCallback } from "react";
+import React, { ComponentProps, useCallback, useMemo } from "react";
 import { GestureResponderEvent, Pressable, View } from "react-native";
 import Animated, {
   Extrapolate,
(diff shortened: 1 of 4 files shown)


src/components/listitems/ListItemInfoCopy.tsx
```diff
@@ -30,9 +30,11 @@ export type ListItemInfoCopy = WithTestID<{
...
   onPress: (event: GestureResponderEvent) => void;
   icon?: IOIcons;
-  // Accessibility
-  accessibilityLabel: string;
-}>;
+}> &
+  Pick<
+    ComponentProps<typeof Pressable>,
+    "accessibilityLabel" | "accessibilityHint"
+  >;
```
CrisTofani, line 37: I would leave the label here as mandatory, since we should highlight that this button aims to copy the value and the I18n value of the "label" would not be available here but in the Main App I18n config. What do you think?

CrisTofani (dismissed):

src/components/listitems/ListItemInfoCopy.tsx
CrisTofani: ok after further analysis we may leave the copy instruction as a hint, specified by the main app

CrisTofani (approved):

PR #327, opened 2024-08-28
[IOPID-2107] - A11Y TextInputBase accessibility improvement

ChrisMattew:
## Short description
Added `accessibilityHint` when prop `status` is equal to "error", improve `TextInputValidation`'s error announce feature.

- **TextInputBase**: added `accessibilityHint` property
- **TextInputValidation**: added possibility to customize error message on input validation, in case of custom validation error messages

## How to test
Using a physical device, enable the screen reader and target an input with an error status; its error message should be announced. Now, in the case of dynamic error messages, it’s possible to pass them through the validation function in order to announce them properly.


src/components/textInput/TextInputBase.tsx
```diff
@@ -392,6 +392,7 @@ export const TextInputBase = ({
           disableFullscreenUI={true}
           accessibilityState={{ disabled }}
           accessibilityLabel={accessibilityLabel ?? placeholder}
+          accessibilityHint={status === 'error' ? bottomMessage : ''}
```
CrisTofani, line 395: I think that assuming this as a generic rule may be lead to some inconsistencies... what about adding another prop to accessibility options and handle any kind of logic on the parent component?

CrisTofani (changes requested):
new validation logic sounds ok, anyway I have some concerns

src/components/textInput/TextInputBase.tsx
ChrisMattew: I completely agree! In this case this logic can be moved in the TextInputValidation component. Excluding from it the accessibilityHint prop from external props. What do you think about it?

CrisTofani: I would leave the problem to define the hint in the component/screen integrating the `TextInputValidation` avoiding to create a generic scope that may lead to wrong behaviours, maybe just define a `accessibilityHint` prop to be drilled down to the `TextInputBase` what do you think?

ChrisMattew: It makes sense! I will implement it!

CrisTofani (approved):

PR #343, opened 2024-10-25
[IOAPPX-412] Add dark mode support to `Badge` and `Tag` components

dmnplb:
## Short description
This PR adds support for dark mode to the `Badge` and `Tag` components

- Add dark mode support for both components
- Add a specific `forceLightMode` prop to `Tag` components to force light mode, even in dark mode, because the component is also used inside the Wallet cards, which probably won't have a dark mode version in the near future

## How to test
Launch the example app and check every component in the **Badges & Tags**


LeleDallas (approved):
Tested positively 
| Dark | Light |
|--------|--------|
| Screenshot 2024-11-20 at 18 00 54 | Screenshot 2024-11-20 at 18 01 19
 |

PR #344, opened 2024-10-28
[IOAPPX-413] Add dark mode support to `Alert`, `Banner` and `FeatureInfo` components

dmnplb:
## Short description
This PR adds support for dark mode to the `Alert`, `Banner` and `FeatureInfo`  components.

## How to test
Launch the example app and go to:
- **Alert** screen
- **Advice & Banners** screen


src/components/banner/Banner.tsx
```diff
@@ -157,9 +157,29 @@ export const Banner = ({
...
   accessibilityLabel,
-  testID,
+  testID
 }: Banner) => {
-  const isPressed: Animated.SharedValue<number> = useSharedValue(0);
+  const isPressed: SharedValue<number> = useSharedValue(0);
+
+  const { isExperimental } = useIOExperimentalDesign();
+  const { themeType } = useIOThemeContext();
+  const theme = useIOTheme();
```
LeleDallas, line 166: How about creating an `isDark` constant derived from the `useIOTheme` hook?

dmnplb: @LeleDallas It's a great suggestion, and it could make the code clearer and less verbose. But it also requires a bit of refactoring on both the `io-app-design-system` and `io-app` codebases, as the `useIOTheme` only returns a single object and we need to return two things: the `theme` object and the `isDark` boolean value.

I'd open a specific PR just for this. What do you think?

LeleDallas: Sounds good to me!

LeleDallas (approved):

PR #345, opened 2024-10-29
[IOAPPX-415] Remove the legacy UI + Add the new `useIONewTypeface` hook to manage new and legacy typography

dmnplb:
## Short description
This PR completely removes the legacy UI by removing all the logic introduced to manage the new and old views simultaneously. It also adds the new `useIONewTypeface` hook to manage the legacy and new typography.

- Remove the legacy styles from all the components
- Replace `isExperimental` with `newTypefaceEnabled` in all the typographic styles

## How to test
1. Launch the example app
2. You should no longer need to activate the experimental DS
3. Test the components by enabling/disabling the new typeface toggle
4. Check that all the components are OK


Ladirico (approved):
For me, this PR is fine, so LGTM! (Good job!) But it is better to wait for [this PR](https://github.com/pagopa/io-app/pull/6422) to be merged on master in app IO before merging this one on DS. I wouldn't want any overlapping. 
In addition, I would add a description if it is to be tested in the app [on this branch](https://github.com/pagopa/io-app/pull/6422). 
Be careful with the alignment to master. If you need a hand I can review you again after the alignment 🚀

PR #346, opened 2024-11-03
[IOAPPX-369] Remove carried over dependencies

LazyAfternoons:
> [!WARNING]  
> Breaking changes to `ListenItemTransaction` component

## Short description
This PR removes several dependencies that were initially included when components were copied from `io-app`. These dependencies in my opinion should not be part of this library as they either perform tasks that can be handled directly with plain JavaScript or implement logic that should be managed within the consuming app.

- Remove `io-ts`, `fp-ts`, `i18n-js` and `@pagopa/ts-commons` dependencies;
- Refactor `LogoPaymentWithFallback.tsx` and `findFirstCaseInsensitive` logic without `fp-ts`;
- Refactor `ListenItemTransaction.tsx` without `fp-ts` and add the `transactionAmountAccessibilityLabel` instead of using `getAccessibleAmountText` which uses `i18n`. Also remove this function;
- Refactor `Tag.tsx` without `fp-ts`;
- Refactor `IODSExperimentalContextProvider.tsx` and `IOThemeContextProvider.tsx` without `fp-ts`;
- Refactors `addCacheTimestampToUri` without `fp-ts` and add tests. `ModuleIDP` and `Avatar` are affected by this change. I had to remove the `undefined` change, fully trusting typescript on this one to not cause a regression on `Avatar` and avoid changing more stuff;
- (Extra) Group up props for `ListItemTransaction` and `ModulePaymentNotice`.

## How to test
Check for any regression in the aforementioned components.


src/components/listitems/ListItemTransaction.tsx
```diff
@@ -76,11 +72,13 @@ export type ListItemTransaction = WithTestID<
           transactionStatus: ListItemTransactionStatusWithoutBadge;
           badgeText?: string;
           transactionAmount: string;
+          transactionAmountAccessibilityLabel: string;
```
dmnplb, line 75: Since we are going to introduce breaking changes, how about grouping the values that are related to the 'transaction' by defining a single object?
```
transaction: {
  amount: string;
  accessibilityLabel: string;
  status: ListItemTransactionStatusWithoutBadge;
}
```

src/components/modules/ModulePaymentNotice.tsx
```diff
@@ -54,6 +55,7 @@ const ModulePaymentNoticeContent = ({
   subtitle,
   paymentNoticeStatus,
   paymentNoticeAmount,
+  paymentNoticeAmountAccessibilityLabel,
```
dmnplb, line 58: As proposed above:
```
paymentNotice: {
  status: PaymentNoticeStatus;
  amount: string;
  accessibilityLabel: string;
}
```

Hantex9 (approved):
LGTM!

dmnplb (commented):
I'd like to propose a change in the component API. @Hantex9 Check them and let me know if it makes sense 👀 

(Does the `@types/react-native-i18n` package still make sense now that we have removed `i18n`?)

Hantex9:
> I'd like to propose a change in the component API. @Hantex9 Check them and let me know if it makes sense 👀
> 
> (Does the `@types/react-native-i18n` package still make sense now that we have removed `i18n`?)

I think that's ok but this means that before we align the io-app with these changes, we need to adapt those elements. I will take care to adapt those elements with your proposed changes ASAP the PR is released.

LazyAfternoons:
> I'd like to propose a change in the component API. @Hantex9 Check them and let me know if it makes sense 👀
> 
> (Does the `@types/react-native-i18n` package still make sense now that we have removed `i18n`?)

I remove a few more dependencies and refactored a bit more code, please refer to commit [7f68757](https://github.com/pagopa/io-app-design-system/pull/346/commits/7f687570cd67978138a728a57763d8ebca0800ce). 

I also updated the props as you suggested, however I specified that the accessibility label refers to the amount, thus calling it `amountAccessibilityLabel`. Refer to commit [d48b398](https://github.com/pagopa/io-app-design-system/pull/346/commits/d48b39811f7d51ed7bc06cd3d6915cdbd8ce5f4c). 

I also updated the PR description accordingly with the new changes. Please test the components and check for any regression.

Hantex9 (approved):

PR #351, opened 2024-11-08
[IOPID-2426] Tooltip component

ChrisMattew:
## Short description
This PR adds the `Tooltip` component

- Created `Tooltip` component
- Added `Tooltips` screen in the Example app to show the `Tooltip` behaviors
## Demo
|iOS|Android|
|----|-------|
|<video src="https://github.com/user-attachments/assets/1060426b-0ec6-4340-b029-6fcdf07cfe10"></video>|<video src="https://github.com/user-attachments/assets/8a76daf8-687c-498e-a54d-ca161fb4a5d0"></video>|

|A11Y iOS|A11Y Android|
|----|-------|
|<video src="https://github.com/user-attachments/assets/a03ee6be-5575-4634-a18a-35079ffa9f08"></video>|<video src="https://github.com/user-attachments/assets/2043f05a-6543-42f5-be14-cb091362e87f"></video>|

>[!Important]
> The `left` and `right` positions are calculated after the component is rendered. This implementation causes a minor glitch where the Tooltip repositions itself to achieve a centered alignment relative to the enclosing element.
> To avoid this behavior, the opacity value of the component depends on the parameters returned by the `onLayout` prop. Once the `tooltipLayout` is defined, the opacity is set to a value of 1.

## How to test
Run the example app and test the `Tooltip` different implementations


thisisjp:
Awesome! Why don't we add a very subtle fade in \ fade out transition to gently darken the background? @dmnplb

dmnplb (approved):
As @thisisjp rightly points out, the component feels a bit static in terms of interaction: we could improve both the background and the entry/exit transitions. But I also want to be pragmatic: I think that the first iteration is good enough to release.

Let's not forget that this component is special because it needs to handle different placements and device configurations, so we should integrate it into the main app as soon as possible to do the appropriate viability tests before making any other improvements.

So, in short: **LGTM**, but let's plan the following refinements:
- Background and entry/exit smooth transitions, ideally made with `reanimated`
- Dark mode support by using the appropriate theme keys

src/components/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,266 @@
...
+  useEffect(() => {
+    if (isVisible) {
+      // A new measure is executed every time the `Tooltip` is visible
+      // This is required for use within ScrollView components.
+      setTimeout(measureChildrenCoords, 100);
+    } else {
+      setChildrenCoords(INITIAL_COORDS);
+      setCurrentPlacement(initialPlacement);
+    }
+  }, [isVisible, initialPlacement, measureChildrenCoords]);
```
Ladirico, line 140: I think it's better to add clearTimeout, what do you think?
```suggestion
 useEffect(() => {
    // eslint-disable-next-line functional/no-let
    let timeoutId: string | number | NodeJS.Timeout | undefined;
    if (isVisible) {
      timeoutId = setTimeout(measureChildrenCoords, 100);
    } else {
      setChildrenCoords(INITIAL_COORDS);
      setCurrentPlacement(initialPlacement);
    }
    return () => clearTimeout(timeoutId);
  }, [isVisible, initialPlacement, measureChildrenCoords]);
```

src/components/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,266 @@
...
+  Modal,
+  Dimensions,
+  LayoutChangeEvent,
+  TouchableWithoutFeedback
+} from "react-native";
+import { useSafeAreaInsets } from "react-native-safe-area-context";
+import { every, some } from "lodash";
+import { IOColors } from '../../core';
+import { Body, H6 } from '../typography';
+import { IconButton } from '../buttons';
```
Ladirico, line 21:
```suggestion
import { IOColors } from "../../core";
import { Body, H6 } from "../typography";
import { IconButton } from "../buttons";
```

src/components/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,266 @@
...
+};
+type CloseWithBackgroundTapDisabled = {
+  allowCloseOnBackgroundTap?: false;
+};
+type Props = CommonProps & (CloseWithTapOnBackground | CloseWithBackgroundTapDisabled);
+
+/**
+ * This component prompts a tooltip around its children in relation to the given `placement`.
+ * When `isVisible` is `true`
+ */
```
Ladirico, line 82: I think it is better to add all the information about the props that the component accepts so that it is clear to people who use it. I have included a proposal for you but feel free to rework it

```suggestion
/**
 * Tooltip component that displays a contextual tooltip around its children.
 * The tooltip position is controlled by the `placement` prop and can adjust
 * dynamically if there is insufficient space.
 *
 * @param {React.ReactNode} children - The element around which the tooltip will be displayed.
 * @param {string} title - The title text displayed at the top of the tooltip.
 * @param {string} content - The main content text of the tooltip.
 * @param {Placement} [placement="top"] - Initial position of the tooltip; can be 'top', 'bottom', 'left', or 'right'.
 * @param {string} closeIconAccessibilityLabel - Accessibility label for the close icon button.
 * @param {boolean} isVisible - Controls the visibility of the tooltip.
 * @param {Partial<DisplayInsets>} [displayInsets={}] - Insets for adjusting tooltip position within screen boundaries.
 * @param {boolean} [allowCloseOnBackgroundTap=false] - Allows closing the tooltip by tapping outside of it.
 * @param {boolean} [childrenInteractionsEnabled=true] - Controls whether the tooltip interacts with accessibility tools.
 * @param {() => void} onClose - Callback function triggered when the tooltip is closed.
 *
 * @returns {React.ReactElement} A tooltip component rendered around the specified children.
 */
```

Ladirico (changes requested):

src/components/tooltip/Tooltip.tsx
ChrisMattew: Addressed [here](https://github.com/pagopa/io-app-design-system/pull/351/commits/3320826cce913626aaeedbfa4201d8420d5b9942)

Ladirico (approved):
LGTM!

PR #353, opened 2024-11-12
[IOAPPX-421] Make `ListItemTransaction` independent of transaction status

dmnplb:
## Short description
This PR removes the specific `status` prop related to `ListItemTransaction`, as this kind of logic should be applied in the main `io-app` repository. Making it agnostic has also improved the code readability.

- Remove `status` and relative types from `ListItemTransaction`
- Add these new props to the `transaction` configuration object:
  - `badge`: set the text and color variant of the badge (if set, `amount` will be hidden)
  - `refund` if you need to signal that the amount is being added and not subtracted
- Rename the `hasChevronRight` to `showChevron`. If `true`, the text will be rendered with the primary interactive colour

## How to test
Go to the **List Items** screen in the example app. Nothing should be changed visually.


src/components/listitems/ListItemTransaction.tsx
```diff
@@ -200,19 +159,29 @@ export const ListItemTransaction = ({
...
-              size={IOListItemVisualParams.chevronSize}
-            />
-          )}
-        </View>
-      </>
-    );
-  };
+      </HStack>
+      <HStack style={{ alignItems: "center" }}>
+        {hideAmount ? (
```
Hantex9, line 164: Instead of using this additional prop, what do you think about hiding the amount when a badge is provided as input?
```suggestion
        {badge ? (
```

dmnplb: @Hantex9 The refactoring was done incrementally, so I completely ignored this optimisation. Good catch, fixed with commit https://github.com/pagopa/io-app-design-system/pull/353/commits/f58901d45dfd428e85bf534652319fcf713e8643)

Hantex9 (approved):
LGTM!

PR #355, opened 2024-11-13
Remove `HeaderSecondLevel` background opacity animation if `transparent: false`

mastro993:
## Short description
This PR removes the background opacity animation in `HeaderSecondLevel` component if the `transparent` prop is set to `false`

- set `headerBgColorTransparentState` color opacity only if `transparent` prop is set to `true`

## How to test
Open the example app, check that the `HeaderSecondLevel` behaves as expected in every use case.


dmnplb (approved):
It doesn't break any header configuration present in the example app, so LGTM

PR #358, opened 2024-11-14
[IOAPPX-428] Add the new `useScaleAnimation` and `useListItemAnimation` hooks + Add support for the `reducedMotion` a11y setting

dmnplb:
## Short description
This PR improves the overall code quality by removing a lot of repeated code. This change is possible thanks to the new `useScaleAnimation` and `useListItemAnimation` hooks. This PR also adds support for the _**Reduce motion**_ a11y setting.

- `useScaleAnimation` should be used when you want a simple scale effect, like the one applied to `Button…` or `Module…` components
- `useListItemAnimation` should be used when you want to build another `ListItem…` component and you can't use `PressableListItemBase` for various reasons

Both hooks use the `useSpringPressProgressValue` hook at their core.

- Add and export the new `useScaleAnimation` and `useListItemAnimation` hooks
- Rename `IOScaleValues` to `IOScaleEffect`, improve naming of its keys
- Apply the new `useScaleAnimation` hook to the following components:
  - All the buttons
  - `PressableModuleBase`, so all the `Module…` components
  - `Alert…` components
  - `Banner`
  - `TabItem`
  - `NumberButton`
- Apply the new `useListItemAnimation` hook to all the **interactive** `ListItem…` components 
- Add support for the _**Reduce Motion**_ setting: when it's enabled, the scaling effect won't be applied when you press, except for the `Module…` components where it's still applied, but with a less pronounced effect
- Replace `useListItemSpringAnimation` with `useListItemAnimation`
- Replace `useModuleSpringAnimation` with `useScaleAnimation`

### Preview
When `reducedMotion` is enabled, the scale effect won't be applied:

https://github.com/user-attachments/assets/2764ce43-b63b-4fc5-acc7-ff51bccc87fe

## How to test
Go to all the components that use a scale effect and check that there are no regressions.


LeleDallas (commented):
The tests on the iOS simulator look promising. I haven’t noticed significant differences compared to the default behavior. It would be great if someone else could test as well to confirm.

LeleDallas (approved):
Test on real devices, works as expected!

PR #359, opened 2024-11-18
[IOAPPX-429] Enable `allowFontScaling` by default in the `IOText` component

dmnplb:
## Short description
This PR enables `allowFontScaling` by default in the `IOText` component. Previously the default was tied to the state of the `isExperimental` value, so it was only enabled if you enabled the experimental DS.

- Set `allowFontScaling` fallback value to `true` in the `IOText` component 

## How to test
Set larger text on your device and check that the text is dynamic in both standard and experimental views.


LeleDallas:
Tested on Android device and iOS devices and it works properly!

| Android  | iOS |
| ------------- | ------------- |
| <video src="https://github.com/user-attachments/assets/3d352b6a-8662-4860-833d-a51f512a56a5">  | <video src="https://github.com/user-attachments/assets/c26c0492-3b60-4e06-a673-ee7fdda8633e">|

On android I'm facing the same behavior described [here](https://github.com/pagopa/io-app-design-system/pull/348#issuecomment-2521104426).
I believe further investigation is needed to address it

LeleDallas (approved):
As discussed during the meeting, we find this behavior acceptable 💫

PR #372, opened 2024-12-19
Remove some variants of `Badge`

dmnplb:
> [!caution]
> This PR introduces a breaking change. It will be merged along with this PR just before we release the `5.0.0` major:
> * https://github.com/pagopa/io-app-design-system/pull/345

## Short description
This PR removes some variants of `Badge` from the DS.

- Remove some variants
- Update documentation and stories

### Preview

https://github.com/user-attachments/assets/6b5d30d8-73f7-4fbe-81b2-f95c69bf8fde

## How to test
Check the **Badges** page in the example app


src/components/badge/Badge.tsx
```diff
@@ -13,28 +13,29 @@ import {
...
-    | "info"
-    | "warning"
-    | "error"
-    | "success"
-    | "purple"
-    | "lightBlue"
-    | "blue"
-    | "turquoise"
-    | "contrast";
+  variant: "default" | "warning" | "error" | "success" | "cgn" | "highlight";
```
LeleDallas, line 38: Since the `variant` parameter is required, what do you think to initialize its value to `default`?

dmnplb: @LeleDallas The `Badge' component always gives a meaning when used, even when it's set as `default'. I don't remember exactly why I set the props as required, but it probably had to do with the side effect when the props are not set and the component uses the fallback. If we change the fallback, some parts of the application will show a different badge, thus conveying a different meaning. Maybe I'm being too pedantic here, but I prefer to keep it mandatory.

Another reason is that most of these badges are configured through a config object, and it's better to stay strict at the TypeScript level.

PR #379, opened 2025-01-10
[IOAPPX-462] Fix flickering behaviour in the `SearchInput` on iOS

dmnplb:
## Short description
This PR fixes the flickering behaviour in the `SearchInput` on iOS.

- Replace `ButtonLink` with `IOText`
- ~~Slightly refactor the component using `reanimated` Shared Values~~

### Preview

The visual issue before the fix:

https://github.com/user-attachments/assets/27b5458c-0705-4ac8-93e6-2863251c9287

## How to test
1. Launch the example app
2. Go to the **SearchInput** screen, after enabling bolder text on iOS

### DIFF
--- a/src/components/searchInput/SearchInput.tsx
+++ b/src/components/searchInput/SearchInput.tsx
@@ -3,7 +3,8 @@ import React, {
   forwardRef,
   useCallback,
   useImperativeHandle,
-  useRef
+  useRef,
+  useState
 } from "react";
 import {
   ColorValue,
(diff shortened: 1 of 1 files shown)


LeleDallas:
Work as expected but be careful for this edge case 👀

Screenshot 2025-01-10 at 18 41 25

LeleDallas (dismissed):

dmnplb:
@LeleDallas Can you check it again? Just replaced `ButtonLink` with `IOText`. Restored the rest.

LeleDallas:
I found this edge case however it disappear when the app is re-rendering the component

https://github.com/user-attachments/assets/110b4ce5-66ce-46bd-844b-f271a46c60c4

dmnplb:
> I found this edge case however it disappear when the app is re-rendering the component

@LeleDallas It's quite annoying and should be fixed, but it seems like a minor issue compared to the original one. I'd merge the PR anyway, even if this problem wasn't fixed.

Ideally, we should replace this custom component with a native one that's certainly more a11y compliant (the one available in the **Example app → Search (under Screens)**

LeleDallas (approved):

PR #380, opened 2025-01-13
[SIW-1978] Fix a11y issues of list item header and action element

gispada:
## Short description
This PR modifies the accessibility props of `ListItemHeader` to fix how the header is announced. To be more specific:
- The header label is announced independently as a header;
- The optional action item on the right is announced with its own accessibility label, without the header label.

- Made the header view accessible with role header
- Removed accessibility from the outer component view
- Removed concatenation of header label and action label

## How to test
Use a screen reader to announce the list item header, with and without the end element. Ensure each part is announced correctly, in the right order.

### DIFF
--- a/src/components/listitems/ListItemHeader.tsx
+++ b/src/components/listitems/ListItemHeader.tsx
@@ -65,30 +65,19 @@ export const ListItemHeader = ({
   const { dynamicFontScale, spacingScaleMultiplier, hugeFontEnabled } =
     useIOFontDynamicScale();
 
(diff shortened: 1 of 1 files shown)


dmnplb:
@gispada Have you already tested the accessibility on both iOS and Android? If so, have you tested on a real physical device or a virtual one?

gispada:
I tested on both platforms on a real device.

dmnplb (approved):

PR #384, opened 2025-01-24
[IOAPPX-467] Remove legacy color palette

dmnplb:
## Short description
This PR completely removes the legacy colour palette, except for the primary `blue` colour, which is still needed to render the legacy UI in the most appropriate way.

- Remove the legacy color palette
- Rename `blue` to `blue-500`
- Leave the legacy color set with the different shades of `blue-500` (`blue-600` and `blue-50`)
- Update tests and snapshots
- Remove two legacy pictograms that were left for legacy support

## How to test
Run the example app and check if everything is okay


example/src/pages/Pictograms.tsx
```diff
@@ -59,16 +58,14 @@ const sortPictogramSet = (
 
 const filteredIOPictograms = filterPictogramSet(
   {
-    ...IOPictogramsObject,
-    ...IOPictogramsLegacy
+    ...IOPictogramsObject
```
LeleDallas, line 61: What do you think to pass the object without spreading it
```suggestion
const filteredIOPictograms = filterPictogramSet(
  IOPictogramsObject,
  IOPictograms
);
```

dmnplb: Good catch! It made sense with more than two objects to filter. But we can just put the single object when it's just one. Fixed in the following commit → [`d44a226` (#384)](https://github.com/pagopa/io-app-design-system/pull/384/commits/d44a226853bf6ca0841f0aeab9078cb6a22918cf)

LeleDallas (approved):

PR #392, opened 2025-02-05
fix: [IOAPPX-478] Fix `HeaderSecondLevel` issue when variant is set explicitly and dark mode is enabled

dmnplb:
## Short description
This PR fixes an issue that occurred when a certain variant was set to `HeaderSecondLevel` and dark mode was enabled.

### The problem
After the recent introduction of dark mode support for `IconButton`, all icon buttons rendered with dark tones were automatically switched to light tones when dark mode was enabled. This behavior is welcome in most cases, but we have some specific app screens where we set a specific variant (e.g., always light or always dark) regarding the enabled color scheme. 

This is especially common when we set a custom background color. In these cases, if you explicitly set the `HeaderSecondLevel` to dark tones (`neutral` variant), the color was automatically switched to light tones when dark mode was enabled.

> [!note]
> The fix only applies to the new UI. Don't expect the fix to work on the legacy UI

- Add a new `persistentColorMode` prop to the `IconButton`. This prevents the color from switching automatically
- Refactor the inner logic of `HeaderSecondLevel` by applying `persistentColorMode` only when variant is set
- **[extra]** Add support for `FlatList` to the `animatedRef` prop

### Preview

https://github.com/user-attachments/assets/ba20d1d3-d96e-4682-9842-b165abe4558a

## How to test
Edit the props of the different `HeaderSecondLevel` screens included in the example app and test them in both light and dark mode

### DIFF
--- a/example/src/pages/HeaderSecondLevelScreenDiscreteTransitionCustomBg.tsx
+++ b/example/src/pages/HeaderSecondLevelScreenDiscreteTransitionCustomBg.tsx
@@ -6,8 +6,8 @@ import {
   IOVisualCostants,
   VSpacer
 } from "@pagopa/io-app-design-system";
-import { useNavigation } from "@react-navigation/native";
 import { useHeaderHeight } from "@react-navigation/elements";
+import { useNavigation } from "@react-navigation/native";
 import * as React from "react";
 import { useLayoutEffect } from "react";
 import { Alert, View } from "react-native";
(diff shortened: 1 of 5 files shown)


example/src/pages/HeaderSecondLevelScreenDiscreteTransitionCustomBg.tsx
```diff
@@ -23,13 +23,11 @@ export const HeaderSecondLevelScreenDiscreteTransitionCustomBg = () => {
...
     navigation.setOptions({
-      headerTransparent: true,
       header: () => (
         <HeaderSecondLevel
           enableDiscreteTransition
-          transparent
-          variant="contrast"
-          backgroundColor={IOColors["blueIO-500"]}
+          variant="neutral"
+          backgroundColor={IOColors["blueIO-100"]}
```
LeleDallas, line 30: What do you think to find a dynamic approach to determine whether the variant should be `neutral` or `contrast` based on the `backgroundColor` value?

dmnplb: It would be very nice, but in other parts of the screens we set colors by expliciting their value (see CGN category screens, for example), so maybe it's better to leave it as it is now to avoid undesired side effects (like parts of the screen with light tones, others parts with dark ones)

LeleDallas (approved):

PR #407, opened 2025-02-19
feat: [SIW-2040] Collapsible list items

gispada:
## Short description
This PR combines the `AccordionItem` and the `ListItemCheckbox` components to create a collapsible component that displays a list of items that can be selected with a checkbox or are just static.

- Modified `AccordionItem` to support an initial expanded state, a secondary background color, and an optional callback on toggle
- Created `CollapsibleListItems`
- Added tests and examples

## How to test
Check the examples in Components > ListItems.
Check Components > Accordion to ensure no regressions were introduced.

https://github.com/user-attachments/assets/592ca09a-19e8-4c43-bd28-138ebb6f75d1


dmnplb (changes requested):
Hi @gispada,
I'm pretty happy because there is a new component in the library :)

I'm honest, I'm not entirely convinced about editing the current `AccordionItem` to accommodate this new scenario. If we want to keep the components easily maintainable, we should keep things separate, because it's not easy to ensure proper `a11y` support if we have a single component that does A, B, and C.

Also, if we need to change some visual parts of this new collapsible for some reason (font size, etc...), we need to add further complexity to manage the different use cases.

To avoid code repetition, I'd rather try to abstract the code around the animation (something similar was done in the `useListAnimation` hook) to make it easier to create similar components in the future.

Last but not least: why give a way to disable the animation?

What do you think?

gispada:
I have extracted the animation logic into `useAccordionAnimation` (do you think it is consistent with the other animation hooks?) and created a separate accordion component in `CollapsibleListItems`. There is still a bit of duplication, but I think it is acceptable.

Regarding the animation, when the accordion starts expanded I don't think the animation should run, and this is the reason why it can be disabled.

src/hooks/useAccordionAnimation.tsx
```diff
@@ -0,0 +1,106 @@
...
+  useAnimatedStyle,
+  useSharedValue,
+  withSpring
+} from "react-native-reanimated";
+import { IOSpacingScale, IOSpringValues } from "../core";
+
+const accordionBodySpacing: IOSpacingScale = 16;
+
+type Params = {
+  defaultExpanded?: boolean;
```
dmnplb, line 13: Why not just `expanded`? 
```suggestion
  expanded?: boolean;
```

src/components/listitems/CollapsibleListItems.tsx
```diff
@@ -0,0 +1,211 @@
...
+  selectionEnabled?: boolean;
+  /**
+   * The IDs of the selected items, when the component is controlled.
+   */
+  selectedItemIds?: Array<string>;
+  /**
+   * Whether the accordion starts expanded.
+   * @default false
+   */
+  defaultExpanded?: boolean;
```
dmnplb, line 46:
```suggestion
  expanded?: boolean;
```

src/components/listitems/CollapsibleListItems.tsx
```diff
@@ -0,0 +1,211 @@
...
+        <View>
+          <VSpacer size={IOSelectionListItemVisualParams.descriptionMargin} />
+          <BodySmall weight="Regular" color={theme["textBody-tertiary"]}>
+            {description}
+          </BodySmall>
+        </View>
+      )}
+    </View>
+  );
+};
```
dmnplb, line 185: We already have a lot of `ListItem...` components suitable for our different needs. In this case we should use a `ListItemInfo` with a simple key-value pair. Let's discuss this during the design process to understand if it is easily replaceable.

dmnplb (changes requested):
> I have extracted the animation logic into useAccordionAnimation (do you think it is consistent with the other animation hooks?) and created a separate accordion component in CollapsibleListItems. There is still a bit of duplication, but I think it is acceptable.

Well done! I think it's much better now. The duplication can be refactored in a second iteration, but at least we keep the component purposes separate.

As we discussed privately, I think the components now have a clear purpose and this should be reflected in the naming, i.e. less generic.

At this point, I'd also rename the `Accordion` screen to `Collapsible` and move it from the `ListItems` page.

Here below just a suggestion about the API. What do you think?

src/components/listitems/CollapsibleListItems.tsx
gispada: Addressed in 1174593452c92e2039ce91dd775365d06a459aa2

dmnplb (approved):

PR #417, opened 2025-02-27
refactor(Tab): Add prop to exclude padding in container

LeleDallas:
This props is useful for the new IOScrollViewLargeHeaderWithList component

## Short description
This pull request introduces changes to the `TabNavigation` to add an optional `excludePadding` prop. This prop allows the user to control whether padding is applied to the container

- Added an optional `excludePadding` property to the `TabNavigation` type.
- Updated the `contentContainerStyle` to conditionally apply padding based on the `excludePadding` prop.

## How to test
N/A

### DIFF
--- a/src/components/tabs/TabNavigation.tsx
+++ b/src/components/tabs/TabNavigation.tsx
@@ -40,7 +40,7 @@ const TabNavigation = ({
   tabAlignment = "center",
   onItemPress,
   children,
-  includeContentMargins = false
+  includeContentMargins = true
 }: TabNavigation) => {
   const [itemMinWidth, setItemMinWidth] = React.useState<number>(0);
 
--- a/src/components/tabs/TabNavigation.tsx
+++ b/src/components/tabs/TabNavigation.tsx
@@ -23,6 +23,7 @@ type TabNavigation = {
   onItemPress?: (index: number) => void;
   // Tabs
   children: TabNavigationChildren;
+  excludePadding?: boolean;
 };
 
 const itemsJustify: Record<TabAlignment, FlexStyle["justifyContent"]> = {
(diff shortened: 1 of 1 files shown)


src/components/tabs/TabNavigation.tsx
```diff
@@ -37,7 +38,8 @@ const TabNavigation = ({
   selectedIndex,
   tabAlignment = "center",
   onItemPress,
-  children
+  children,
+  excludePadding = false
```
dmnplb, line 42: How about keeping the same convention used for the `IOScrollViewWith...` component?
```suggestion
  includeContentMargins = true
```

src/components/tabs/TabNavigation.tsx
```diff
@@ -79,7 +81,7 @@ const TabNavigation = ({
       centerContent={true}
       showsHorizontalScrollIndicator={false}
       contentContainerStyle={[
-        styles.container,
+        excludePadding ? styles.containerNoPadding : styles.container,
```
dmnplb, line 84: If we change the naming, we can also make the logic more explicit by using a condition without negation:
```suggestion
        includeContentMargins
          ? { paddingHorizontal: IOVisualCostants.appMarginDefault }
          : {},
```
We should also add `flexGrow` in the following style:
```jsx
{
  flexGrow: 1,
  justifyContent: itemsJustify[tabAlignment]
}
```

src/components/tabs/TabNavigation.tsx
```diff
@@ -95,6 +97,9 @@ const styles = StyleSheet.create({
     flexGrow: 1,
     paddingHorizontal: 24
   },
+  containerNoPadding: {
+    flexGrow: 1
+  },
```
dmnplb, line 102: So we can get rid of both `container…` style objects:
```suggestion
```

dmnplb (changes requested):
Here's my suggestions. What do you think?

LeleDallas:
> Here's my suggestions. What do you think?

That's way better! Addressed here https://github.com/pagopa/io-app-design-system/pull/417/commits/9b73ca384d429f47046fb0a4885a10745078abf9

dmnplb (approved):

PR #427, opened 2025-03-11
feat: [IOPLT-966, IOPLT-991] Add the new `IOSkeleton` component to remove the `rn-placeholder` dependency

dmnplb:
## Short description
This PR adds the new `IOSkeleton` component to remove the obsolete and abandoned `rn-placeholder` dependency.

### Features
Here are the features, compared to the legacy dependency:
- The new component supports only the `pulse` animation
- It's based on the `reanimated` library
- It supports the _Reduced motion_ system setting to improve accessibility
  - If it's enabled, the pulse animation is disabled and the blocks are rendered with a static gray shade
- It includes support for dark mode

- Add and export the new `IOSkeleton` component
- Replace the legacy `Placeholder.Box` components with the new `IOSkeleton` component
- Remove `rn-placeholder` dependency from both the library and the example app

### Preview
In the example app you can find a screen with a series of repeated skeletons for testing purposes:

https://github.com/user-attachments/assets/933c42f6-a8be-4f48-a142-63475df02802

## How to test
Run the example app with different devices to see if there are any performance hiccups.


LeleDallas (commented):
Tested with `One+ Android 15`, the transition is very slow:

https://github.com/user-attachments/assets/c2447cb2-d53c-4564-ad51-b68c57fe88ef

LeleDallas:
> @LeleDallas Just updated the example app to use `FlatList` instead of the less performant `map` from the `Array`. Let me know if you notice any improvements, but keep in mind that the example app is stuck at version `0.72`, while the main repo is at `0.75`. I don't know if it really matters, but we should also test in the main repo.
> 
> (cc @CrisTofani)



> @LeleDallas Just updated the example app to use `FlatList` instead of the less performant `map` from the `Array`. Let me know if you notice any improvements, but keep in mind that the example app is stuck at version `0.72`, while the main repo is at `0.75`. I don't know if it really matters, but we should also test in the main repo.
> 
> (cc @CrisTofani)

It's a bit better now, but using `logcat`, I can see this warning:
```
Skipped N frames!  The application may be doing too much work on its main thread.
```

Interestingly, removing two instances of `IOSkeleton` from `Skeleton` seems to resolve the issue. This suggests that rendering multiple `IOSkeleton` components might be overloading the main thread.

CrisTofani (approved):
LGTM

PR #428, opened 2025-03-11
fix(`TextInputBase`): correct counter limit handling

LeleDallas:
## Short description
This pull request fixes the issue where the `counterLimit` was not correctly accounting for the spaces added by the `valueFormat` function in the `TextInputBase` component. This ensures that the actual text length (excluding spaces) is compared against the `counterLimit`.

- Updated the `onChangeTextHandler` to correctly handle the `counterLimit` by calculating the actual text length without spaces
- Adjusted the `maxLength` prop of the `TextInput` component to account for the spaces added by the `valueFormat` function
- Ensured that the `counterLimit` is respected correctly for formatted text inputs such as `credit-card` and `iban`

## How to test
- Set a `counterLimit` in `DSTextFields`
- Insert some value in `credit-card` and `iban` `TextInput`
- Ensure that now the `counterLimit` displaying the correct number
## Preview
| Old | New |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/6449ed43-75fb-4eec-b633-2a018c7e78e8" /> | <video src="https://github.com/user-attachments/assets/0de2d7f0-2b01-4751-b759-a0c41108198e"/>  | 

### Prevent counter updating on wrong value

| Old  | New |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/bb0bab8f-bb4f-4433-aa10-f2ed1e87cfa1" /> | <video src="https://github.com/user-attachments/assets/7518a037-e9d4-4b48-aba8-2877b2d71e90"/>  |

### DIFF
--- a/src/components/textInput/TextInputBase.tsx
+++ b/src/components/textInput/TextInputBase.tsx
@@ -120,17 +120,27 @@ const styles = StyleSheet.create({
 
 type InputTextHelperRow = Pick<
   InputTextProps,
-  "value" | "counterLimit" | "bottomMessage" | "bottomMessageColor"
+  | "value"
+  | "counterLimit"
+  | "bottomMessage"
+  | "bottomMessageColor"
+  | "inputType"
 >;
 
 const HelperRow = ({
   value,
   counterLimit,
   bottomMessage,
-  bottomMessageColor
+  bottomMessageColor,
+  inputType
 }: InputTextHelperRow) => {
   const theme = useIOTheme();
-  const valueCount = useMemo(() => value.length, [value]);
+
+  const valueCount = useMemo(
+    () =>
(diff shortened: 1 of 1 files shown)


src/components/textInput/TextInputBase.tsx
```diff
@@ -315,12 +315,21 @@ export const TextInputBase = ({
 
   const onChangeTextHandler = useCallback(
     (text: string) => {
-      if (counterLimit && text.length > counterLimit) {
+      const actualTextLength = text.replace(/\s/g, "").length;
```
CrisTofani, line 318: I guess this replacement may not be always applied, what happens if the user needs to input a text with space and we're omitting from the count?

CrisTofani (approved):

CrisTofani (approved):

PR #432, opened 2025-03-17
fix: [IOPLT-1028] Add alternative `accessibilityRole` to `Banner`, `FeatureInfo` and all the `asLink` typographic styles

dmnplb:
## Short description
This PR adds an alternative `accessibilityRole` to `Banner`, `FeatureInfo` and all the typographic styles with the `asLink` prop.

- Add possible alternative `button` accessibility role to all the typographic styles with the `asLink` prop
- Add possibile alternative `button` accessibility role to `FeatureInfo`
- Add possibile alternative `link` accessibility role to `Banner`

## How to test
Run the example app and try to edit these components with custom `accessibilityRole`

### DIFF
--- a/src/components/banner/Banner.tsx
+++ b/src/components/banner/Banner.tsx
@@ -79,9 +79,9 @@ type RequiredBannerProps =
 
 type BannerActionProps =
   | {
-      action?: string;
+      action: string;
       onPress: (event: GestureResponderEvent) => void;
-      accessibilityRole?: never;
+      accessibilityRole?: Extract<AccessibilityRole, "button" | "link">;
     }
   | {
       action?: never;
@@ -142,6 +142,7 @@ export const Banner = ({
   onClose,
   accessibilityHint,
   accessibilityLabel,
+  accessibilityRole = "button",
   testID
 }: Banner) => {
   const { newTypefaceEnabled } = useIONewTypeface();
@@ -178,7 +179,7 @@ export const Banner = ({
         // A11y related props
         accessibilityLabel={accessibilityLabel ?? fallbackAccessibilityLabel}
         accessibilityHint={accessibilityHint}
-        accessibilityRole={action !== undefined ? "button" : undefined}
+        accessibilityRole={action !== undefined ? accessibilityRole : "text"}
       >
         {title && (
           <>
@@ -264,9 +265,6 @@ export const Banner = ({
       style={[styles.container, dynamicContainerStyles]}
       // A11y related props
       accessible={false}
-      accessibilityHint={accessibilityHint}
-      accessibilityLabel={accessibilityLabel}
-      accessibilityRole={"text"}
     >
       {renderMainBlock()}
     </View>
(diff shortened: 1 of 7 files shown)


src/components/banner/Banner.tsx
```diff
@@ -178,7 +179,7 @@ export const Banner = ({
         // A11y related props
         accessibilityLabel={accessibilityLabel ?? fallbackAccessibilityLabel}
         accessibilityHint={accessibilityHint}
-        accessibilityRole={action !== undefined ? "button" : undefined}
+        accessibilityRole={action !== undefined ? accessibilityRole : "text"}
```
CrisTofani, line 182: Would't it be better to leave an `undefined` role instead of `text`?

src/components/typography/Body.tsx
```diff
@@ -24,7 +24,13 @@ const legacyFontName: IOFontFamily = "TitilliumSansPro";
  */
 export const Body = forwardRef<View, BodyStyleProps>(
   (
-    { weight: customWeight, color: customColor, asLink, ...props },
+    {
+      weight: customWeight,
+      color: customColor,
+      asLink,
+      accessibilityRole = "link",
```
CrisTofani, line 31: If we define "link" as a default value if `asLink` is false the component would have link as role anyway, so maybe we should leave it undefined and handle the undefined case only when the `asLink` is true, what do you think?

src/components/typography/BodySmall.tsx
```diff
@@ -21,7 +21,13 @@ const legacyFontName: IOFontFamily = "TitilliumSansPro";
  */
 export const BodySmall = forwardRef<View, BodySmallProps>(
   (
-    { weight: customWeight, color: customColor, asLink, ...props },
+    {
+      weight: customWeight,
+      color: customColor,
+      asLink,
+      accessibilityRole = "link",
```
CrisTofani, line 28: Same as `Body`

src/components/typography/LabelMini.tsx
```diff
@@ -21,7 +21,13 @@ const legacyFontName: IOFontFamily = "TitilliumSansPro";
  */
 export const LabelMini = forwardRef<View, LabelMiniProps>(
   (
-    { weight: customWeight, color: customColor, asLink, ...props },
+    {
+      weight: customWeight,
+      color: customColor,
+      asLink,
+      accessibilityRole = "link",
```
CrisTofani, line 28: same ad `Body`

CrisTofani (changes requested):

src/components/typography/Body.tsx
dmnplb: That's not quite true, because we only pass the `accessibilityRole` prop when `asLink` is `true`. If you look at the generated snapshots, nothing really changes.

Does it make sense?

CrisTofani: yeah correct, I missed the ternary on IOText, that's correct!

ChrisMattew (approved):
LGTM

PR #434, opened 2025-03-18
chore: adjust `ModuleCheckout` a11y label

LeleDallas:
## Short description
This pull request adds an `accessibilityLabel` to the `ModuleCheckout` component to improve screen reader interaction, ensuring that the `CTA` text is announced correctly on Android, which was previously missing.

- Add an `accessibilityLabel` to the `PressableModuleBase` component to enhance screen reader support by including the `title` and `ctaText` properties.

## How to test
- Enable TalkBack on Android or VoiceOver on iOS
- Navigate over a `ModuleCheckout` component
- Ensure that both Android and iOS correctly announce the `text` and `cta` if present

### DIFF
--- a/src/components/modules/ModuleCheckout.tsx
+++ b/src/components/modules/ModuleCheckout.tsx
@@ -84,7 +84,10 @@ export const ModuleCheckout = (props: ModuleCheckoutProps) => {
   );
 
   return ctaText ? (
-    <PressableModuleBase onPress={onPress}>
+    <PressableModuleBase
+      onPress={onPress}
+      accessibilityLabel={`${title} , ${ctaText}`}
+    >
       <HStack space={4} style={{ alignItems: "center" }}>
         <ModuleBaseContent />
         <View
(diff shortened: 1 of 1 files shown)


dmnplb (dismissed):

src/components/modules/ModuleCheckout.tsx
```diff
@@ -84,7 +84,10 @@ export const ModuleCheckout = (props: ModuleCheckoutProps) => {
   );
 
   return ctaText ? (
-    <PressableModuleBase onPress={onPress}>
+    <PressableModuleBase
+      onPress={onPress}
+      accessibilityLabel={`${title} , ${ctaText}`}
```
dmnplb, line 89: What if I need to listen to the details of the payment method, such as the value of the email address associated with it? Looking at the available examples, we can set the `subtitle` optional field, but it's completely ignored with this template string configuration. What do you think?

dmnplb (changes requested):

dmnplb (approved):

PR #441, opened 2025-03-21
chore: [IOPLT-971] Improve accessibility and performance of the loaders + Add support for dark mode

dmnplb:
## Short description
This PR improves accessibility and performance of the loaders and adds support for dark mode.

- **`LoadingSpinner`**
  - Add dark mode support
  - Partial refactoring to add a cleanup function to the animation to prevent memory leaks
  - Remove color set in `Module…` and `ListItem…` components because it's already baked in the original component
- **`ProgressLoader`**
  - Add dark mode support
  - Add `accessibilityRole` and `accessibilityValue` to improve the accessibility of the component
  

### Preview

https://github.com/user-attachments/assets/76dbe347-3f4a-474a-ac28-46acb345cf49

## How to test
Run the example app. Go to the **Loaders** page. Enable/Disable dark mode for testing purposes

### DIFF
--- a/example/src/pages/Loaders.tsx
+++ b/example/src/pages/Loaders.tsx
@@ -2,17 +2,18 @@ import {
   H3,
   IOColors,
   LoadingSpinner,
-  ProgressLoader
+  ProgressLoader,
+  VStack
 } from "@pagopa/io-app-design-system";
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { View } from "react-native";
 import { Screen } from "../components/Screen";
 
 export const Loaders = () => {
-  const [progress, setProgress] = React.useState(0);
-  React.useEffect(() => {
+  const [progress, setProgress] = useState(0);
+
+  useEffect(() => {
     const interval = setInterval(() => {
-      // console.log("progress", progress, (progress + 10) % 100);
       setProgress(prev => (prev + 10) % 100);
     }, 500);
     return () => clearInterval(interval);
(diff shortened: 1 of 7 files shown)


LeleDallas (changes requested):
In Android I see that the spinner are not accessible and the `ProgressLoader` component is announcing it's percentage without a focus

https://github.com/user-attachments/assets/0db2fd26-cfb7-4a10-a717-81126cae4420

On iOS this works properly:

https://github.com/user-attachments/assets/3cd348b9-a79a-4037-9c0b-b564a845ae51

dmnplb:
> In Android I see that the spinner are not accessible and the `ProgressLoader` component is announcing it's percentage without a focus

Addressed here → https://github.com/pagopa/io-app-design-system/pull/441/commits/5ddb9e576e89989b52e95c83b72365c19d499675

On Android, the focus ring didn't appear because the component was too thin. I added a fictional minimum height to make it visible. I also tried setting `accessibilityLiveRegion` to `polite` to communicate updates of the value, but the SR kept reading the updated value even when I had focus on other components. I preferred to disable it.

LeleDallas:
> > In Android I see that the spinner are not accessible and the `ProgressLoader` component is announcing it's percentage without a focus
> 
> @LeleDallas Addressed here → [5ddb9e5](https://github.com/pagopa/io-app-design-system/commit/5ddb9e576e89989b52e95c83b72365c19d499675)
> 
> On Android, the focus ring didn't appear because the component was too thin. I added a fictional minimum height to make it visible. I also tried setting `accessibilityLiveRegion` to `polite` to communicate updates of the value, but the SR kept reading the updated value even when I had focus on other components. I preferred to disable it.

I think you miss an `accessible` flag in `LoadingSpinner → Animated.View` component. Without that flag on Android spinners are still ureachable

dmnplb:
> I think you miss an accessible flag in LoadingSpinner → Animated.View component. Without that flag on Android spinners are still ureachable

Are you sure? I tested the behavior on the real device and the focus ring was placed as expected. It should say "Progress Bar" in the absence of `accessibilityLabel` 🤔

LeleDallas:
> > I think you miss an accessible flag in LoadingSpinner → Animated.View component. Without that flag on Android spinners are still ureachable
> 
> Are you sure? I tested the behavior on the real device and the focus ring was placed as expected. It should say "Progress Bar" in the absence of `accessibilityLabel` 🤔

Was a cache problem, my bad

LeleDallas (approved):

PR #447, opened 2025-03-28
fix: [IOPLT-1049] Refactor `Avatar` and `AvatarSearch` to restore the correct behavior

dmnplb:
## Short description
This PR refactors `Avatar` and `AvatarSearch` to correctly handle all possible scenarios. In particular, `AvatarSearch` caused a crash in the example app due to missing `width` and `height` parameters in the `source` prop. `AvatarSearch` also incorrectly handled the case of multiple elements in the array, where the first is a broken asset, but the second is not.

- Update `Avatar` and `AvatarSearch` components
- Update `Logos` page with a simplified mock object
- Update the inner logic of `Avatar` because the array check is already included in the `addCacheTimestampToUri` function

## How to test
Check `Avatar` and `AvatarSearch` in the `Logos` page


LeleDallas (commented):
Now the application is not throwing an error but this warning:
> [!WARNING]
>  ReactImageView: Image source "null" doesn't exist

dmnplb:
> Now the application is not throwing an error but this warning

@LeleDallas Forgot to mention in the PR description. From RN docs:
> [!Note]
> Note: On Android, the default source prop is ignored on debug builds.

If you try to access the page from an iOS device, you shouldn't see any warning. My conclusion is that Android **only throws this warning on debug builds** because it can't load the asset set in the `defaultSource` prop. What do you think?

LeleDallas:
> > Now the application is not throwing an error but this warning
> 
> @LeleDallas Forgot to mention in the PR description. From RN docs:
> 
> Note
> 
> Note: On Android, the default source prop is ignored on debug builds.
> 
> If you try to access the page from an iOS device, you shouldn't see any warning. My conclusion is that Android **only throws this warning on debug builds** because it can't load the asset set in the `defaultSource` prop. What do you think?

I agree with that statement

PR #448, opened 2025-03-28
fix: [IOPLT-933] Fix CLI error when you try to install the library on external repositories

dmnplb:
## Short description
This PR fixes a CLI error when you try to install the library on external repositories. The error is caused by a lack of support for 3rd party plugins, which we used to add a `postinstall` hook for local development.

## Checks before the merge
- Make sure that the package is properly built (`bob build`) before publishing it to the NPM registry
  - `yarn` v1.x runs `postInstall`, while `yarn` v3.x doesn't

- Remove `postInstall` hook
- Add `yarn prepack` to automatic checks to spot possible errors before the merge

## How to test
Point `io-app` to this specific PR, then run `yarn`. If everything is okay, the `yarn` command shouldn't generate errors.


CrisTofani (approved):
Tested on io-app repo, LGTM

PR #449, opened 2025-03-31
fix: [IOPLT-1049] Restore previous `AvatarSearch` logic to avoid unexpected behavior in search results

dmnplb:
## Short description
This PR restores previous `AvatarSearch` logic to avoid unexpected behavior in search results, due to `FlashList` optimisation techniques.

- Remove `useState` and `onError` function

## How to test
For testing purposes, please refer to the previous merged PR:
- https://github.com/pagopa/io-app-design-system/pull/447

### DIFF
--- a/src/components/avatar/Avatar.tsx
+++ b/src/components/avatar/Avatar.tsx
@@ -149,21 +149,6 @@ export const AvatarSearch = React.memo(
     const internalSpace = dimensionsMap.small.internalSpace;
     const innerRadius = borderRadius - internalSpace;
 
-    const indexValue = useRef<number>(0);
-    const [imageSource, setImageSource] = useState(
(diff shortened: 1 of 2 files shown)


LeleDallas (changes requested):
- iOS: works fine
- Android: some logos are not rendering

https://github.com/user-attachments/assets/6186f55a-a1d8-47f0-bea3-faeeae1b55cc

dmnplb:
> Android: some logos are not rendering

@LeleDallas Sorry if I didn't write it explicitly, but it works as expected. We removed the fallback logic because using `useState` in a component with `FlashList` is not recommended and could cause some unexpected behaviour (like messing up the logos). Compared to the previous iteration, other logos don't load properly because they are part of a data object that is intentionally wrong.

LeleDallas (approved):

PR #453, opened 2025-04-03
feat: [SIW-2201]  Add optional description to ListItemHeader

RiccardoMolinari95:
## Short description
This PR add an optional `description` to `ListItemHeader` component

- Add optional description if present
- Add to example app a `ListItemHeader` with a description

## How to test
Check in the example app the `ListItemHeader` with the description

### DIFF
--- a/example/src/pages/Buttons.tsx
+++ b/example/src/pages/Buttons.tsx
@@ -690,9 +690,7 @@ export const Buttons = () => (
           accessibilityLabel="Go down"
           accessibilityHint="Tap to trigger test alert"
           icon="arrowBottom"
-          onPress={() => {
-            alert("Action triggered");
-          }}
+          onPress={onButtonPress}
         />
 
         <HSpacer size={16} />
@@ -703,9 +701,7 @@ export const Buttons = () => (
           accessibilityHint="Tap to trigger test alert"
           icon="arrowBottom"
           disabled
-          onPress={() => {
-            alert("Action triggered");
-          }}
+          onPress={onButtonPress}
         />
       </View>
     </ComponentViewerBox>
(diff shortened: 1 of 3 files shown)


dmnplb (changes requested):
Add a new `ListItemChekbox` instance close to the updated `ListItemHeader`, just to see if both can stay on the same page without noticeable differences.

We cannot change the colour of the text, but we can change the alignment of the icons:

| Before | After |
|--------|--------|
| Simulator Screenshot - iPhone 16 Pro - 2025-04-03 at 10 41 13 | Simulator Screenshot - iPhone 16 Pro - 2025-04-03 at 10 42 30 | 

To obtain this alignment, we just need to set `iconMargin` to `IOVisualCostants.iconMargin` instead of leave it at `8`.

What do you think?

RiccardoMolinari95:
> Add a new `ListItemChekbox` instance close to the updated `ListItemHeader`, just to see if both can stay on the same page without noticeable differences.
> 
> We cannot change the colour of the text, but we can change the alignment of the icons:
> 
> Before	After
> Simulator Screenshot - iPhone 16 Pro - 2025-04-03 at 10 41 13	Simulator Screenshot - iPhone 16 Pro - 2025-04-03 at 10 42 30
> To obtain this alignment, we just need to set `iconMargin` to `IOVisualCostants.iconMargin` instead of leave it at `8`.
> 
> What do you think?

Thank you @dmnplb , I hadn't noticed this detail. Addressed in [e487c6d](https://github.com/pagopa/io-app-design-system/pull/453/commits/e487c6d6af391e2ba468d87b11aacfd4b201ade1) I used `IOVisualConstants.iconMargin` as you suggested

PR #477, opened 2025-05-23
fix: [IOPLT-1152] Set `maxFontSizeMultiplier` to standard Text Inputs

dmnplb:
## Short description
This PR sets a `maxFontSizeMultiplier` to standard Text Inputs. The issue was silent until now because there was a specific property that generally disabled the `allowFontScaling` property on all the `TextInput` components in the `io-app` codebase.

## How to test
Increase the font size to the maximum and check if the text input is completely broken or usable.

### DIFF
--- a/src/components/textInput/TextInputBase.tsx
+++ b/src/components/textInput/TextInputBase.tsx
@@ -35,7 +35,11 @@ import {
   useIOTheme
 } from "../../core";
 import { useIOFontDynamicScale } from "../../utils/accessibility";
-import { IOFontSize, makeFontStyleObject } from "../../utils/fonts";
+import {
+  IOFontSize,
+  IOMaxFontSizeMultiplier,
+  makeFontStyleObject
+} from "../../utils/fonts";
 import { RNTextInputProps, getInputPropsByType } from "../../utils/textInput";
 import { InputType, WithTestID } from "../../utils/types";
 import { IOIconSizeScale, IOIcons, Icon } from "../icons";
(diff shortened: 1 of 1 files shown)


src/components/textInput/TextInputBase.tsx
```diff
@@ -464,6 +468,7 @@ export const TextInputBase = ({
           selectionColor={IOColors[theme["interactiveElem-default"]]} // Caret on iOS
           cursorColor={IOColors[theme["interactiveElem-default"]]} // Caret Android
           maxLength={adjustedMaxLength}
+          maxFontSizeMultiplier={IOMaxFontSizeMultiplier}
```
shadowsheep1, line 471: What do you think about allowing it to be passed from the outside as well, instead of relying solely on a default?

src/components/textInput/TextInputBase.tsx
```diff
@@ -511,6 +516,7 @@ export const TextInputBase = ({
             onLayout={getLabelWidth}
             numberOfLines={1}
             accessible={false}
+            maxFontSizeMultiplier={IOMaxFontSizeMultiplier}
```
shadowsheep1, line 519: As above

shadowsheep1 (changes requested):

src/components/textInput/TextInputBase.tsx
dmnplb: @shadowsheep1 Unless we need to set a specific value, I'd leave it as it is. The `IOMaxFontSizeMultiplier` value was set based on an analysis of user preferences. For example, `1.2` is too low for some of our citizens.

While I understand that it could be useful in some cases, if we start adding exceptions, the experience for people with low vision will become inconsistent.

shadowsheep1 (approved):
LGTM

PR #483, opened 2025-07-03
fix: [IOPLT-1172] Add a `Pressable` wrapper to typographic components when `asLink` is enabled

dmnplb:
## Short description
This PR addresses an accessibility issue that arises when typographic components are rendered as links. These components cannot be properly handled by keyboards or generic external devices if they are not wrapped in explicitly interactive components.

- Return a wrapped typographic component if `asLink` is set to `true`


LeleDallas (approved):
Tested with switch access on Android and works as expected!

https://github.com/user-attachments/assets/dcd34345-9af5-4cdb-9eb6-5fc63153bbf8

I also verified with voice over functionality

Ladirico (approved):
LGTM!

PR #489, opened 2025-09-02
feat: [SIW-2827] Add the new `ModuleNavigationAlt` component 

MonicaRungi-EY:
## Short description
This PR introduces a new component called `ModuleNavigationAlt` and a new icon called `securityPad`

- Added the new `ModuleNavigationAlt` component, which aligns with the new identification design. The component also accepts a React element as the _image_ prop to include a custom SVG from io-app. Additionally, the right icon has been made customizable.

- Added the new icon `IconSecurityPad`
- Updated `package.json` and `metro.config.js`: to test the visibility of _Spid extended icon_, `react-native-svg-transformer` has been added to _devDependencies_, with the proper configuration in `metro.config.js`

## How to test
Run the example app and go to Components section, then check che Modules screen. The new component, along with the new icon, can be found at the end of the screen.

> ❗️update: Spid Icon
The _Spid Extended icon_ does not respect the theme of the app, this is just a test to check the behavior of the SVG.

## Preview
Screenshot 2025-09-03 alle 11 36 59


src/components/modules/ModuleNavigationAlt.tsx
```diff
@@ -0,0 +1,156 @@
...
+            </BodySmall>
+            {subtitle && (
+              <LabelMini weight="Regular" color={theme["textBody-tertiary"]}>
+                {subtitle}
+              </LabelMini>
+            )}
+          </View>
+        </HStack>
+        {onPress ? (
+          <Icon
```
ale-mazz, line 113: What do you think about making this icon customizable? We could consider passing it from outside, so that it can be customized in case of different future versions.

dmnplb (changes requested):
It's a good place to start.

I've just noticed some things we should consider before merging:

- The main icon on the left doesn't change colour when dark mode is enabled, so we should add the relevant theme key, as with the other inner elements of the component
- According to the design, we should have a fetching spinner component like the other modules. After adding the support for this prop, we should list it on the relevant page in the example app for documentation purposes.
- I'd try a different combination, for example the _Cie_ or _Spid extended icon_, to see if the component still needs refining. If we are unsure how it will behave, we will need to open a new PR just for that detail.

PR #556, opened 2026-03-27
chore: [IOPLT-1727] Fix local environment for example app

dmnplb:
## Short description
Fix local environment for example app.

- Update `.ruby-version` to avoid generating a different `Podfile.lock`
- Update `Podfile` to avoid failed build with XCode 26.4 and above


example/ios/Podfile
```diff
@@ -31,4 +31,21 @@ target 'ExampleApp' do
       # :ccache_enabled => true
     )
   end
-end
+
+  # Workaround for fmt compilation error with Xcode 26.4 / newer Apple Clang
+  # https://github.com/expo/expo/issues/44229
```
LeleDallas, line 36: It may be helpful to include a TODO indicating that this code should be removed once the related PR introduces a proper solution.
What do you think?

dmnplb: I added a simple TODO, nothing very detailed, but at least it signals that it's a temporary hack

LeleDallas (approved):
I was able to run the app with Xcode 26.4!
What do you think to apply the same workaround to the IO App?

dmnplb:
> What do you think to apply the same workaround to the IO App?

I'm already using it locally, but given the number of contributors to the main app, I'm less keen to add it.

PR #559, opened 2026-04-14
fix: [SIW-3657] Replace pause icon

DavideOnano:
## Short description
This PR delete old pause icon and upload new stop icon from DS
- Added stop icon to icons set

## How to test
Verify the icon is coherent with DS


dmnplb (changes requested):
If we consider that the new icon represents `stop` rather than `pause`, does it make sense to rename it from `IconPause` to `IconStop` and also change the relevant key?

dmnplb (approved):

PR #561, opened 2026-04-27
chore: [IOPLT-1800] Replace `forwardRef` with just `ref` prop

dmnplb:
## Short description
Replace `forwardRef` with just the `ref` prop, by taking advantage of the new React version (`19.x`).

> [!caution]
> This PR introduces a potential breaking change because it requires React 19.x or later

- Simplify the logic by removing `forwardRef`
- Exclude `example` from the library's type check and `eslint` config to avoid conflicts

## How to test
Run the app, try to add a specific `ref` prop in the example app and check if there are any errors


LeleDallas (changes requested):
Opening some files, I see this error inside the IDE:
```
Parsing error: ESLint was configured to run on `<tsconfigRootDir>/example/src/pages/Alert.tsx` using `parserOptions.project`: <tsconfigRootDir>/tsconfig.json
However, that TSConfig does not include this file. Either:
- Change ESLint's list of included files to not include this file
- Change that TSConfig to include this file
- Create a new TSConfig that includes this file and include it in your parserOptions.project
See the typescript-eslint docs for more info: https://tseslint.com/none-of-those-tsconfigs-include-this-file
```
In order to solve it, add this line `tsconfigRootDir: __dirname` inside example [parserOptions](https://github.com/pagopa/io-app-design-system/blob/main/example/.eslintrc.js#L12)

dmnplb:
> In order to solve it, add this line `tsconfigRootDir: __dirname` inside example [parserOptions](https://github.com/pagopa/io-app-design-system/blob/main/example/.eslintrc.js#L12)

@LeleDallas Addressed here: https://github.com/pagopa/io-app-design-system/pull/561/changes/9059bcceb2e6a0f7fca8ce8c928d5f52d5d3405b

example/.eslintrc.js
LeleDallas, line 1: Still have some issues:
```
Definition for rule '@typescript-eslint/member-delimiter-style' was not found.
```
Cutting out `@typescript-eslint/member-delimiter-style`, `semi` and `@typescript-eslint/semi` rules seems to solve it

dmnplb: Addressed here → https://github.com/pagopa/io-app-design-system/pull/561/commits/fb4adcf3be83d597530b3432952b469a1fded447

LeleDallas (approved):

PR #562, opened 2026-05-07
chore:[SIW-3657] Enhance accessibility for OTP input component

DavideOnano:
## Short description
Improve OTP Input accessibility properties
- Separate `AccessibiltyValue` for a better experience while TalkBack and VoiceOver reading the secret value
- Avoid `announceForAccessibility` for secret values so is not repeated more times
## Android
Test the OTP Input using TalkBack and ensure the Input is focusable, has a value/label and a role, and that no regression are introduced
## iOS
N.B: I didn't run any test with VoiceOver, please ensure the behavior is correct.
Check with the Accessibility Inspector if the `Input` has a role, value and the correct label.

P.S: By adding `value.split("").join(", "))` we prevents the value from being read as a single number the first time, which appears to be the default behavior of OTPInput. Instead, each digit is read separately.
This is only a workaround, though. I couldn’t find a proper solution for this improvement, and I think it’s outside the scope of the current task.

If you have any suggestions, we can include it in this PR.


dmnplb (changes requested):
Thanks for the proposed improvement.

Just two notes:
- When `secret` is `true` and no `accessibilityValueText` is provided, the input becomes completely silent (empty value text + suppressed announcements): a screen-reader user gets no feedback that anything was entered or how many digits remain.  So the right solution is to make `accessibilityValueText` mandatory when `secret` is used and enforce this at type level via a discriminated union (we can't add an hardcoded string to the library because the system is agnostic). This should highlight all the uncovered cases.
- The Android `opacity: 0` → `0.01` change is undocumented

DavideOnano:
> Thanks for the proposed improvement.
> 
> Just two notes:
> 
> * When `secret` is `true` and no `accessibilityValueText` is provided, the input becomes completely silent (empty value text + suppressed announcements): a screen-reader user gets no feedback that anything was entered or how many digits remain.  So the right solution is to make `accessibilityValueText` mandatory when `secret` is used and enforce this at type level via a discriminated union (we can't add an hardcoded string to the library because the system is agnostic). This should highlight all the uncovered cases.
> * The Android `opacity: 0` → `0.01` change is undocumented

Thank you for your feedback.
I added a discriminated union so now the accessibility value is mandatory if secret is true.
I also remove the platform check for opacity set since we are making the same change both for android and iOS and documented it.
Let me know what do you think!

example/src/pages/OTPInput.tsx
```diff
@@ -16,26 +16,42 @@ const OTP_LENGTH_6 = 6;
...
+  (
+    | {
+        secret: true;
+        accessibilityValueText: AccessibilityValueText;
+      }
+    | {
+        secret?: false;
+        accessibilityValueText?: AccessibilityValueText;
+      }
+  );
```
dmnplb: Redeclaring the types is a code smell. Instead, let's export the entire type from the main component so that it can be easily imported in future without the risk of discrepancies between the props.

src/components/otpInput/OTPInput.tsx
```diff
@@ -126,11 +138,24 @@ export const OTPInput = ({
...
     }
   };
 
+  const accessibilityValueTextParams = {
+    valueLength: value.length,
+    length
+  };
+
+  const accessibilityValueText =
+    props.secret === true
```
dmnplb: We already declare the const by assigning the value based on props, no need to repeat it
```suggestion
    isSecret
```

dmnplb (changes requested):

dmnplb (approved):
After the applied changes, LGTM

DavideOnano:
@dmnplb Thank you! I can't directly merge it tho.
If we don't need a second approval can you do it for me?

PR #564, opened 2026-05-11
refactor: [IEL-000] Show counter limit prop

LeleDallas:
## Short description
This pull request introduces a new feature to the `TextInputBase` and `TextInputValidation` components that allows the character counter to be shown only when the character limit is reached. 

> [!IMPORTANT]
>  The goal of this development is to fix a bug where the `counterLimit` prop is initially set to `undefined` and later updated to a numeric value once a specific condition is met

- Added the `showCounterOnlyWhenLimitReached` prop to `InputTextProps` and `TextInputValidationProps`, allowing the character counter to be displayed only when the limit is reached
- Updated the `HelperRow` component and its logic to use the new prop, so the counter is only shown when appropriate

## How to test
In IO app point to this [development](https://github.com/pagopa/io-app/pull/8093) and ensure that is possible to edit input form when counter limit is reached

## Preview

https://github.com/user-attachments/assets/81b1bf92-1a0d-414f-b009-3507cbc4c197

### DIFF
--- a/src/components/textInput/TextInputBase.tsx
+++ b/src/components/textInput/TextInputBase.tsx
@@ -55,6 +55,7 @@ type InputTextProps = WithTestID<{
   icon?: IOIcons;
   rightElement?: ReactNode;
   counterLimit?: number;
+  showCounterOnlyWhenLimitReached?: boolean;
   accessibilityAnnounceLimitReached?: string;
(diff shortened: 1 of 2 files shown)


Ladirico (approved):
LGTM

